PR #8, opened 2016-07-25
Resums

tiffanyqi:
Convert total time calculators to include start and end times (doesn't work)


MaxFangX:
Will be working on this stuff this week, and I think I'd like to take a slightly different approach. Closing

Issue #30, opened 2016-09-11
Add, edit, and delete tags


tiffanyqi:

MaxFangX:
I don't think issues are automatically closed unless you reference them in the pull request **description**. Ran into this issue with my last PR.

PR #43, opened 2016-10-16
Add tags

tiffanyqi:
- Allow users to add, edit, and remove tags in AJAX.
- Better tag display

Note: 
- Also deleted some spaces in models.py
- Not sure how to remove static/.DS_Store from this PR. Thoughts?


tiffanyqi:
#30

.gitignore
```diff
@@ -16,6 +16,7 @@ cal/cal/static/bower_components/*
 
 # DS Store
 .DS_Store
+cal/cal/static/.DS_Store
```
MaxFangX: You can change line 18 to `*.DS_Store` and  I believe it should work

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
 analyticsApp.controller('LoggedInCtrl', function LoggedInController($scope) {
 });
 
-// Controller to generate graph data from chart directive, cumulative tags
 analyticsApp.controller('TagsCtrl', function($scope, $http){
   var url = '/v1/tags.json';
+  var tagUrl = '/v1/tags/';
+
   $scope.url = url;
```
MaxFangX: Not sure why I added this before, but could you delete it? From a cursory glance it doesn't look like this is being used

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
 analyticsApp.controller('LoggedInCtrl', function LoggedInController($scope) {
 });
 
-// Controller to generate graph data from chart directive, cumulative tags
 analyticsApp.controller('TagsCtrl', function($scope, $http){
   var url = '/v1/tags.json';
+  var tagUrl = '/v1/tags/';
```
MaxFangX: Hmm, so we now have both `/v1/tags/`(1) and `/v1/tags.json`(2).

I think the cleanest way to write these is something like
`tagUrl = "/v1/tags";`
and in the $http in line 14 do
`tagUrl + ".json"`
and whenever we need to access the TagDetail endpoint do
`tagUrl + "/[id]"`

Note that `/v1/tags` is the same as `/v1/tags/` so probably better to leave the last slash out. Thoughts?

MaxFangX: Actually, is `/v1/tags/.json` the same as `/v1/tags.json`? If so, then I think it would be better to have `tagUrl` be `/v1/tags/`. so that we don't have to do `tagUrl + '/' + id`

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+            id: data.id,
+            label: data.label,
+            keywords: data.keywords,
+            hours: data.hours,
+            showEdit: false
+          });
+        });
+    };
+
+    this.editTag = function(tagId) {
```
MaxFangX: How about `startEdit`?

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+          id: tag.id,
           label: tag.label,
           keywords: tag.keywords,
           hours: tag.hours
         });
       }
-
     });
 
+    this.addTag = function(tag) {
```
MaxFangX: How about `create`?

We don't need to namespace by including Tag in the function name since we're binding to the TagsCtrl instead of to the scope (because we're doing `this.XXX` instead of `$scope.XXX`

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+        }
+      }).
+        success(function editToList(data) {
+          tag.label = data.label;
+          tag.keywords = data.keywords;
+          tag.hours = data.hours;
+        });
+    };
+
+    this.cancel = function(tagId) {
```
MaxFangX: How about `cancelEdit`?

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+      var tag = search(tagId, $scope.tags);
+      tag.showEdit = false;
+    };
+
+    this.deleteTag = function(tagId) {
```
MaxFangX: `delete`

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+        }
+      }).
+        success(function removeTagFromList(data) {
+          $scope.tags = $scope.tags.filter(function(tag) {
+            return tag.id !== tagId;
+          });
+        });
+    };
+
+    function search(id, tags){
```
MaxFangX: This would be a good place to use a library method instead of writing our own (similar to how you use $scope.tags.filter a few lines before). You can do `$scope.tags.find(function ...)`, which is native to Javascript arrays. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

cal/cal/templates/home_logged_in.html
```diff
@@ -80,13 +81,53 @@ <h4 class="well">Categories - Last Week</h4>
...
-      </ul>
+        <table class="table table-hover">
+          <thead>
+            <th>Label</th>
+            <th>Hours</th>
+          </thead>
+          <tbody>
+            <tr ng-repeat="tag in tags">
+              {% verbatim %}
+              <td ng-show="!tag.showEdit">{{ tag.label }}</td>
```
MaxFangX: More semantically correct to say `ng-hide="tag.editing"`. This occurs a few other times as well

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+          'Content-Type': 'application/x-www-form-urlencoded'
+        }
+      }).
+        success(function addToList(data) {
+          $scope.tags.push({
+            id: data.id,
+            label: data.label,
+            keywords: data.keywords,
+            hours: data.hours,
+            showEdit: false
```
MaxFangX: Let's use `editing` instead of showEdit so that it makes more sense with the `ng-show` and `ng-hide` directives.

cal/cal/templates/home_logged_in.html
```diff
@@ -80,13 +81,53 @@ <h4 class="well">Categories - Last Week</h4>
...
+              <td ng-show="!tag.showEdit">{{ tag.hours }}</td>
+              {% endverbatim %}
+              <td ng-show="tag.showEdit">
+                Label<br><input type="text" ng-model="tag.new_label">
```
MaxFangX: Can we have the input on the next line, since we're using a `<br>`? Happens a few other places as well

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+          });
+        });
+    };
+
+    this.editTag = function(tagId) {
+      var tag = search(tagId, $scope.tags);
+      tag.new_label = tag.label;
```
MaxFangX: New label and keywords should be `camelCase`

cal/cal/templates/home_logged_in.html
```diff
@@ -80,13 +81,53 @@ <h4 class="well">Categories - Last Week</h4>
...
+              </td>
+              <td ng-show="tag.showEdit">
+                Keywords<br><input type="text" ng-model="tag.new_keywords">
```
MaxFangX: camelCase

cal/cal/templates/home_logged_in.html
```diff
@@ -80,13 +81,53 @@ <h4 class="well">Categories - Last Week</h4>
...
+                  <span class="glyphicon glyphicon-edit" ng-click="$ctrl.editTag(tag.id)"></span>
+                </button>
+                <button class="btn btn-danger">
+                  <span class="glyphicon glyphicon-trash" ng-click="$ctrl.deleteTag(tag.id)"></span>
+                </button>
+              </td>
+            </tr>
+            <tr>
+              <td>Label <br> <input type="text" ng-model="tag.label"></td>
+              <td>Keywords <br> <input type="text" ng-model="tag.keywords"></td>
```
MaxFangX: THere is a space after label and keywords

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
         });
       }
-
     });
 
+    this.addTag = function(tag) {
+      $http({
+        method: 'POST',
+        url: url,
+        data: $.param({
```
MaxFangX: I believe you can use `params: {...}` instead of mixing in jQuery with `data: $.param({`. See the arguments section in [here](https://docs.angularjs.org/api/ng/service/$http)

cal/cal/static/js/app.js
```diff
@@ -3,32 +3,120 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+          label: tag.new_label,
+          keywords: tag.new_keywords,
+          csrfmiddlewaretoken: getCookie('csrftoken'),
+          _method: 'PATCH'
+        }),
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded'
+        }
+      }).
+        success(function editToList(data) {
```
MaxFangX: How about `addToList`?

MaxFangX (changes requested):

cal/cal/static/js/app.js
tiffanyqi: @MaxFangX just checked, `v1/tags/.json` throws an error. would you like me to change it to the first comment?

cal/cal/static/js/app.js
tiffanyqi: @MaxFangX you mean `$scope.url = url`?

cal/cal/static/js/app.js
tiffanyqi: changing to `create`

.gitignore
tiffanyqi: @MaxFangX changed this line, it's still showing up as a file change

cal/cal/static/js/app.js
MaxFangX: Sounds good!

cal/cal/static/js/app.js
MaxFangX: Yes

MaxFangX:
I looked at your DS_Store thing again. `*.DS_Store` looks like the correct rule as per the reference .gitignore by Git itself: [https://github.com/github/gitignore/blob/afbff9027d02ccfc680e031f6c295f79ad61662d/Global/macOS.gitignore#L1](https://github.com/github/gitignore/blob/afbff9027d02ccfc680e031f6c295f79ad61662d/Global/macOS.gitignore#L1)

cal/cal/static/js/app.js
tiffanyqi: @MaxFangX python doesn't like: `  $http({method: 'GET', url: tagUrl + '.json' }).`, any suggestions?

tiffanyqi:
@MaxFangX should we just leave it on the commit?

MaxFangX:
Fine by me

cal/cal/static/js/app.js
tiffanyqi: changed to:

```
    function search(id, tags){
      return tags.find(function (element, index, array) {
        return element.id == id;
      });
    }
```

cal/cal/static/js/app.js
tiffanyqi: fixed! (ignore deleted comment)

cal/cal/static/js/app.js
tiffanyqi: @MaxFangX not really understand the documentation you've provided me. How do I change the query parameters into a successful request? This is the last part for me.

cal/cal/static/js/app.js
MaxFangX: Ahh, I should have been more clear. I meant that every time you're calling search, it should be implemented with a `.find(` instead. For example in like 83 it should be:

`var tag = $scope.tags.find(function(tag, index, array) { return tag.id == id; });`

Which probably seems verbose since you have to write that everywhere you're using currently using search, but is better than implementing a library function ourselves.

The reason we have to do this is because we haven't finished issue #41 yet, which would change our array accesses to constant time by using a dictionary instead of an array.

MaxFangX: If you're feeling like it, all we need to do to implement #41 is to change our `$scope.tags` object to a dictionary where the keys are tag ids and the values are the tag objects (it's not that hard, you can totally do it!). See [here](https://stackoverflow.com/questions/1208222/how-do-i-implement-a-dictionary-or-hashtable-in-javascript). Then you can access tags quickly and non-verbosely with `$scope.tags[id]` instead of what I wrote in my previous comment. Just be careful that you make the necessary modifications to support that, including:
- Making sure our `ng-repeat` iterates over the values of the dictionary
- Ensuring that all additions/deletions/accesses/modifications to our dictionary use the correct dictionary syntax.

I hope that is enough information, let me know if you have any questions!

MaxFangX: [Relevant stack overflow](https://stackoverflow.com/questions/11985863/how-to-use-ng-repeat-for-dictionaries-in-angularjs)

cal/cal/static/js/app.js
MaxFangX:
```
url: url,
params: {
    label: tag.label,
    keywords: tag.keywords, 
    csrfmiddlewaretoken: ... 
},
headers: ...
```

It that doesn't work then dwbi

tiffanyqi: tried this, didn't work, will leave as is

cal/cal/static/js/app.js
tiffanyqi: @MaxFangX `var tag = $scope.tags.find(function(tag, index, array) { return tag.id == id; });` to me seems like copying and pasting every time when we can just use one function that does call this called `search`. wouldn't this still be constant time since we're still using the `find` library function?

MaxFangX: `find` is linear time because it has to search through an array. It does exactly what you had previously implemented in your `search` function, so changing `search` to make a call to `find` is redundant.

tiffanyqi:
@MaxFangX changed everything you requested. Please check this out when you get the chance!

MaxFangX (approved):

Issue #31, opened 2016-09-12
Multiple stages of auth need to be consolidated - PSA and Google OAuth


MaxFangX:

tiffanyqi:
@MaxFangX is this complete with your latest update?

MaxFangX:
@tiffanyqi Yes! Issue should close once PR is merged

PR #48, opened 2016-10-18
Fix auth

MaxFangX:
- Sends the login access_token to backend when signing into the app
- Saves the access_token
- When requesting scopes for Google OAuth2, will bundle new permissions with previously granted permissions and store in the same access_token
- When logging in, if OAuth2 (calendar) permissions were granted previously, the user will skip the second stage of auth and retain a valid access_token for syncing
- Some changes to signup flow, such as associating users with an email address provided by Google


MaxFangX:
Fixes #31 (We have agreed on incremental authorization, referenced issue is no longer relevant)

cal/cal/views.py
```diff
@@ -46,6 +44,27 @@ def accounts_profile(request):
...
+    user = request.backend.do_auth(token)
+    if user:
+        login(request, user)
+
+        # The user agent is only used for logs
+        credential = AccessTokenCredentials(token, 'dummy-user-agent/1.0')
+        storage = Storage(GoogleCredentials, 'user', request.user, 'credential')
+        storage.put(credential)
+
+        return HttpResponseRedirect("/")
```
tiffanyqi: Why not say:

```
if
...
return HttpResponseRedirect("/")
```

instead of returning the same thing twice?

MaxFangX: Done!

tiffanyqi (approved):
LGTM

PR #57, opened 2016-10-24
Default event colors use calendar color, move all day events to top of UI

### LINKED ISSUE #25: Show colors of events accurately
If people have separate calendars: use the calendar color rather than the event color for default event colors

MaxFangX:
Fixes #25 
- Adds new field `color_index` to GCalendar, pulls `colorId` info from google calendar API
- Display all_day_events as such on frontend ui
- Use calendar color when event color is default
- Refactor `import_calendars` a little bit

Make sure you run migrations after pulling
After you've done so, run this in the shell:

```
my_user = User.objects.get(username="myusername")
my_user.googlecredentials.import_calendars(False)
```

This command will update your calendars with the correct color index.


cal/cal/migrations/0043_gcalendar_color_index.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('cal', '0042_auto_20160925_0730'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='gcalendar',
+            name='color_index',
+            field=models.CharField(default='10', max_length=10, choices=[(b'1', b'#a4bdfc'), (b'2', b'#7ae7bf'), (b'3', b'#dbadff'), (b'4', b'#ff887c'), (b'5', b'#fbd75b'), (b'6', b'#ffb878'), (b'7', b'#46d6db'), (b'8', b'#e1e1e1'), (b'9', b'#5484ed'), (b'10', b'#51b749'), (b'11', b'#dc2127')]),
```
tiffanyqi, line 17: What are these? Are these the google calendar colors, and where are you using this in the code?

cal/cal/models.py
```diff
@@ -717,17 +736,19 @@ def import_calendars(self, only_primary=True):
...
                 profile.main_calendar = gcal
                 profile.save()
-            elif not only_primary:
-                gcal, gcal_created = GCalendar.objects.get_or_create(user=self.user, calendar_id=item['id'])
-                if gcal_created:
-                    gcal.update_meta()
+
+            # Update fields for each calendar
+            gcal.summary = item.get('summary')
+            gcal.color_index = item.get('colorId', '1')
```
tiffanyqi, line 749: what's the '1'? is that the default color for the calendar? if so, why is it applied to all?

cal/cal/static/js/app.js
```diff
@@ -189,6 +189,7 @@ analyticsApp.controller('CalendarCtrl', function UiCalendarCtrl($scope, $http, $
                           title: gevent.name,
                           start: gevent.start,
                           end: gevent.end,
+                          allDay: gevent.all_day_event,
```
tiffanyqi, line 192: +1 tested, this works so well

tiffanyqi (changes requested):

tiffanyqi (approved):

tiffanyqi:
Oops, made a mistake. Meant to "comment" but not review, but then approved my changes (oops). Could you address the first two comments please? Thanks!

cal/cal/migrations/0043_gcalendar_color_index.py
MaxFangX: This is the generated migrations code from Django. It adds a new field color_index with all the possible options of colors as defined in constants.py. There are list comprehensions in the GCalendar class that you may look at in the shell

MaxFangX: It is used in the code in GEvent.color when it refers to self.calrndar.color!

cal/cal/models.py
MaxFangX: get avoids key errors by returning a default value if a specified key is not found. '1' is the default color, yes, and will be set as such when a colorId key doesn't exist

tiffanyqi:
@MaxFangX thanks! LGTM. Waiting on @jalagar, if he'd like to take a look

jamesuejio:
+2

Issue #29, opened 2016-09-11
Re-add, delete, and edit names of categories


tiffanyqi:
Note: there will be a "generate categories" button in which users will automatically see all categories that's they have on their calendar, which includes different iterations of colored events and calendars.
- Re-add a deleted category.
- Delete a category from view.
- Edit the names of categories.

Issue #59, opened 2016-10-24
Deleting calendars produces UnicodeEncodeError at /admin/cal/gcalendar/


tiffanyqi:
Ways to reproduce:
1. Go to /admin
2. Select a calendar to delete

See error:
UnicodeEncodeError at /admin/cal/gcalendar/
'ascii' codec can't encode character u'\u2019' in position 18: ordinal not in range(128)

```
Traceback:
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/contrib/admin/options.py" in wrapper
  616.                 return self.admin_site.admin_view(view)(*args, **kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/utils/decorators.py" in _wrapped_view
  110.                     response = view_func(request, *args, **kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/views/decorators/cache.py" in _wrapped_view_func
  57.         response = view_func(request, *args, **kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/contrib/admin/sites.py" in inner
  233.             return view(request, *args, **kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/utils/decorators.py" in _wrapper
  34.             return bound_func(*args, **kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/utils/decorators.py" in _wrapped_view
  110.                     response = view_func(request, *args, **kwargs)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/utils/decorators.py" in bound_func
  30.                 return func.__get__(self, type(self))(*args2, **kwargs2)
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/contrib/admin/options.py" in changelist_view
  1574.                 response = self.response_action(request, queryset=cl.get_queryset(request))
File "/Users/Tiffany1/Desktop/temp/calenv/lib/python2.7/site-packages/django/contrib/admin/options.py" in response_action
[...]

PR #61, opened 2016-10-24
Edit, Delete, and Sort Categories

tiffanyqi:
#29
- Add a new API endpoint to access any given category
- Dynamic textbox resizing (when editing or adding tags and categories)
- Users can now check/uncheck categories, which connects to the pie chart
- Users can now edit label names and delete categories
- Remove label name from the chart itself
- Remove extraneous spaces and make lines consistent


cal/cal/static/css/calendaranalytics.scss
```diff
@@ -25,6 +25,12 @@ button {
   padding: 0 5 5 5;
 }
 
+.textbox input {
+  width:100%;
+  box-sizing:border-box;
+  -moz-box-sizing:border-box;
```
MaxFangX: Spaces after the colons please! To match the other rules/ the standard

cal/cal/static/js/app.js
```diff
@@ -102,33 +102,125 @@ analyticsApp.controller('TagsCtrl', function($scope, $http){
...
+          label: category.label,
+          hours: category.hours,
+          include: true
+        });
+        $scope.categoriesForChart.push({
+          id: category.id,
           label: category.label,
-          hours: category.hours
+          hours: category.hours,
+          include: true
```
MaxFangX: Here would be a good place to set a temporary variable representing the category javascript object instead of explicitly setting the same keys and values twice

cal/cal/static/js/app.js
```diff
@@ -102,33 +102,125 @@ analyticsApp.controller('TagsCtrl', function($scope, $http){
...
+    });
+
+    // when user checks a category
+    if (category.include) {
+      $scope.categoriesForChart.push({
+        id: category.id,
+        label: category.label,
+        hours: category.hours,
+      });
+      category.include = true;
```
MaxFangX: I believe this is supposed to be `= false`?

cal/cal/static/js/app.js
```diff
@@ -102,33 +102,125 @@ analyticsApp.controller('TagsCtrl', function($scope, $http){
...
+
+    // when user unchecks a category
+    } else {
+      $scope.categoriesForChart = $scope.categoriesForChart.filter(function(category) {
+        return category.id !== categoryId;
+      });
+      category.include = false;
```
MaxFangX: `= true`?

cal/cal/templates/home_logged_in.html
```diff
@@ -72,15 +72,54 @@
...
+              </td>
+            </tr>
+            <tr>
+              <td class="text-nowrap">
+                <button class="btn btn-success">Generate Categories</button>
+              </td>
+            </tr>
+          </tbody>
+        </table>
+      <nvd3 options="options" data="categoriesForChart"></nvd3>
```
MaxFangX: Instead of maintaining a whole new array of categories in memory just for the chart, you can use a `filter` angular filter by `include` here. [Link to docs](https://docs.angularjs.org/api/ng/filter/filter)

cal/cal/static/js/app.js
MaxFangX: Actually, you can ignore this. See my comment about `filter`

cal/api/serializers.py
```diff
@@ -59,14 +60,15 @@ def get_fields(self, *args, **kwargs):
...
-        fields = ('color_index', 'label', 'hours', 'calendar')
+        fields = ('id', 'color_index', 'label', 'hours', 'calendar')
 
     def get_hours(self, obj):
-        calendar_ids = self.context['calendar_ids']
-        start = self.context['start']
-        end = self.context['end']
+        calendar_ids = self.context.get('calendar_ids')
+        start = self.context.get('start')
+        end = self.context.get('end')
```
MaxFangX: Instead of changing these to `get`, you should add the same `get_serializer_context` method from `ColorCategoryList` to the `ColorCategoryDetail` class so that these three keys will always exist in the context dictionary (it's accessed in the serializer's `get_hours` method). This is because we want our API endpoints for single categories to be able to filter by calendar ids, start, and end as well. It's my fault that I missed adding this code for `TagDetail`, so while you're at it, could you add the same code to `TagDetail` as well?

cal/cal/static/css/calendaranalytics.scss
```diff
@@ -25,6 +25,12 @@ button {
   padding: 0 5 5 5;
 }
 
+.textbox input {
```
MaxFangX: Instead of .textbox input you can just place the textbox class on the input element itself. We also namespace all of our custom classnames with `cal-foo-bar` to avoid name collisions with bootstrap and other libraries, and to make it obvious that we're not using bootstrap styling (so it should be `cal-textbox`). There are other examples of this in this same scss file.

MaxFangX (changes requested):

cal/cal/static/js/app.js
tiffanyqi: Surprisingly, it should be `true`. I tried switching this around, and it doesn't do uncheck properly.

cal/cal/static/js/app.js
tiffanyqi: Surprisingly, it should be `false`. I tried switching this around, and it doesn't check properly.

cal/cal/static/css/calendaranalytics.scss
tiffanyqi: Is there a particular reason why you want it on `input` instead of `.textbox input`? When I change it to `input`, it puts an extra line in between the checkbox and the label name, which doesn't look particularly good. I'll also change the naming.

cal/api/serializers.py
tiffanyqi: @MaxFangX Copied `get_serializer_context` from `api/views.py` into `ColorCategoryDetail` as requested. However when I test it, it gets this error: `super(type, obj): obj must be an instance or subtype of type`, any idea why?

cal/cal/templates/home_logged_in.html
tiffanyqi: @MaxFangX I'm having a really hard time getting this to work. I've tried:

`nvd3 options="options" data="categories | filter:category:include"` and

```
$scope.getFilteredCategories = $scope.categories.filter(function(category) {
  return category.include;
});
```

both of which do not update the pie chart automatically, unlike the previous solution. Any pointers? Thanks!

tiffanyqi: I'm thinking it's because they're not using AJAX right now, and thus aren't updating the pie chart automatically.

cal/api/serializers.py
MaxFangX: Oh, that's because you need to pass `ColorCategoryDetail` into `super` instead of `ColorCategoryList`. Recall that `super` allows you to call the inherited method that you're overriding, and it doesn't make sense to introduce another class entirely

cal/cal/static/js/app.js
MaxFangX: I don't think this line of code is working the way you expect it to. In line 198 we have an if statement asking if `category.include` is `true`, and since line 204 is in this if block set it to true, it's redundant because `category.include` should already be set to true.

You can delete this whole if-else block though, since we don't need the `categoriesForChart` variable.

cal/cal/static/js/app.js
MaxFangX: See other comment

cal/api/views.py
```diff
@@ -159,6 +159,15 @@ class ColorCategoryDetail(generics.RetrieveUpdateDestroyAPIView):
 
     serializer_class = ColorCategorySerializer
 
+    # def get_serializer_context(self, *args, **kwargs):
+    #     context = super(ColorCategoryList, self).get_serializer_context(*args, **kwargs)
```
MaxFangX: Should be `ColorCategoryDetail`

MaxFangX: Should be `ColorCategoryDetail`

cal/cal/static/js/app.js
```diff
@@ -101,7 +101,7 @@ analyticsApp.controller('TagsCtrl', function($scope, $http){
     };
 });
 
-analyticsApp.controller('CategoriesCtrl', function($scope, $http){
+analyticsApp.controller('CategoriesCtrl', function($scope, $http, $filter){
```
MaxFangX: You don't need to inject `$filter` because you're not using it anywhere in this controller. We only need it in the html, which Angular will pick up automatically

cal/cal/templates/home_logged_in.html
```diff
@@ -117,6 +117,8 @@ <h4 class="well">Categories - Last Week</h4>
             </tr>
           </tbody>
         </table>
+<!--       <nvd3 options="options" data="getFilteredCategories"></nvd3>-->
+<!--       <nvd3 options="options" data="categories | filter:category:include"></nvd3>-->
```
MaxFangX: This isn't working because the `expression` field needs to be a string, Object, or function. I recommend filtering using the Object. Looks like you're using the comparator field wrong too. Full specification in the docs [here](https://docs.angularjs.org/api/ng/filter/filter) (Look at 'arguments'). If you can't figure out the docs let me know!

cal/cal/static/js/app.js
```diff
@@ -124,6 +124,9 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
           include: true
         });
       }
+      // $scope.getFilteredCategories = $scope.categories.filter(function(category) {
+      //   return category.include;
+      // });
```
MaxFangX: You can delete these - we can do all the filtering conveniently in the html

cal/cal/static/js/app.js
```diff
@@ -207,6 +210,9 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
       });
       category.include = false;
     }
+      // $scope.getFilteredCategories = $scope.categories.filter(function(data) {
+      //     return data.include;
+      // });
```
MaxFangX: You can delete these

MaxFangX (changes requested):

cal/cal/templates/home_logged_in.html
```diff
@@ -89,7 +89,7 @@ <h4 class="well">Categories - Last Week</h4>
                 {{ category.hours }}
               </td>
               {% endverbatim %}
-              <td ng-show="category.editing" class="textbox">
+              <td ng-show="category.editing" class="cal-text-box">
```
MaxFangX: These `cal-text-box` classes should be put on the `input` elements.

cal/cal/static/css/calendaranalytics.scss
MaxFangX: Reason is that we're using an extra selector when we don't need one. The styling is only ever needed for the inner element, so we don't need to select the outer element and then reselect the inner element.

cal/cal/static/css/calendaranalytics.css
```diff
@@ -19,6 +19,11 @@ button {
   float: right;
   padding: 0 5 5 5; }
 
+.cal-text-box input {
```
MaxFangX: This should be just `cal-text-box`

jamesuejio:
Generate categories seems to not work. Also I noticed in the tags and categories, adding and deleting can sometimes be glitchy. Also is there any way to get back categories that I deleted? I accidently deleted other, and generate categories doesn't work.

cal/cal/static/css/calendaranalytics.scss
tiffanyqi: Done. My initial issue was that changing the class to be placed on `input` also affected the checkbox input. Changed to `input[type=text]`!

tiffanyqi:
@jalagar this commit does not contain logic on re-adding categories. I feel as though we'd have to wait until generated categories works again. Can you put in / reopen issues on what you experienced? Thanks!

cal/cal/templates/home_logged_in.html
tiffanyqi: @MaxFangX I did this as requested, and turned it into: `<nvd3 options="options" data="categories | filter:{include:true}"></nvd3>`, which works as expected.

However, I am receiving an infinity digest loop. According to [stack overflow](http://stackoverflow.com/questions/20963462/rootscopeinfdig-error-caused-by-filter), this is caused when you are changing the model that you are filtering, and `ng-repeat` checks to see if there are any changes over and over again. The suggested practice is to avoid changing the model in the filter.

While this method does work, the impact of the infinity digest is botching our intended style of the page (all the hidden textboxes and buttons are emerging on the table). Thoughts?

jamesuejio:
@tiffanyqi Wait so should I create an issue that generate categories doesn't work? Also I noticed in app.js, all the tabs got messed up, and it doesn't include my merge with event click functionality. Just wondering if it's going to be a pain to merge into master.

tiffanyqi:
@jalagar I suggest you create an issue detailing the steps you took to get the problem you did, and the stack trace you get back. Take a look at #59 for an example. Generate categories works ok for me, so it'd be good if we could see why yours isn't working!

Thank you for mentioning your onClick method; I'll add it right back in. Did I miss anything else?

MaxFangX (approved):

MaxFangX:
+2

Issue #60, opened 2016-10-24
Sort tables by name or hours


tiffanyqi:
- Sort category table
- Sort tabs table

PR #62, opened 2016-10-25
Tags table sorts by label or hours

tiffanyqi:
#60

Allows the tags table to sort by label or hours. Once categories is merged, this code can easily be copied to categories as well.


cal/cal/static/css/calendaranalytics.scss
```diff
@@ -20,6 +20,10 @@ button {
   margin: 2px;
 }
 
+a[ng-click]{
```
MaxFangX: What is this for? Are you doing this because hovering over a link with ng-click doesn't show the hand? If so, may I suggest you try `href=""`? Generally, using attribute names in css selectors is messy when adding a class will do.

cal/cal/templates/home_logged_in.html
```diff
@@ -85,11 +85,19 @@ <h4 class="well">Categories - Last Week</h4>
...
+            </th>
+            <th>
+              <a ng-click="orderByField='hours'; reverseSort = !reverseSort">
+                Hours
+              </a>
+            </th>
           </thead>
           <tbody>
-            <tr ng-repeat="tag in tags">
+            <tr ng-repeat="tag in tags|orderBy:orderByField:reverseSort">
```
MaxFangX: Don't want to be nitpicky but could you follow the spacing standard set [here](https://docs.angularjs.org/api/ng/filter/orderBy) (spaces before and after the pipe symbol)

```
    </tr>
    <tr ng-repeat="friend in friends | orderBy:'-age'">
      <td>{{friend.name}}</td>
```

MaxFangX (changes requested):

cal/cal/static/css/calendaranalytics.scss
tiffanyqi: This was a hacky way to prevent link clicking to go up to the top of the page. Your solution also combatted that so I amended it, thanks!

cal/cal/templates/home_logged_in.html
tiffanyqi: Changed. Is this better? Thanks!

MaxFangX (approved):

PR #73, opened 2016-10-31
Change Google OAuth include_granted_scopes parameter to string

### LINKED ISSUE #72: Google OAuth include_granted_scopes parameter requires string parameter
Emily and I ran into this issue when setting up the dev environment for the first time. Imo it's pretty stupid but at least they gave us a descriptive error message.

MaxFangX:
Fixes #72


jamesuejio (approved):
Why is this the case? Is it just how it's implemented to use a string instead?

MaxFangX:
Yeah, Google API requires a string `true`. It's throwing an error because we're sending along a capital `True` instead

PR #75, opened 2016-11-06
Change implementation from list to dict.

### LINKED ISSUE #41: Optimize tags and categories - should not have to search through array every time

jamesuejio:
Fixes #41

### DIFF
--- a/cal/cal/templates/home_logged_in.html
+++ b/cal/cal/templates/home_logged_in.html
@@ -98,7 +98,7 @@
             </th>
           </thead>
           <tbody>
-            <tr ng-repeat="tag in tags | orderBy:orderByField:reverseSort">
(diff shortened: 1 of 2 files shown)


cal/cal/templates/home_logged_in.html
```diff
@@ -98,7 +98,7 @@ <h4 class="well">Tags - Cumulative</h4>
             </th>
           </thead>
           <tbody>
-            <tr ng-repeat="tag in tags | orderBy:orderByField:reverseSort">
```
jamesuejio, line 101: Wait to merge until I can figure this out, looks like orderBy doesn't work for dictionaries:
http://stackoverflow.com/questions/14330878/orderby-not-working-with-dict-syntax-on-ng-repeat

PR #76, opened 2016-11-07
New pages

MaxFangX:
- Detail pages for both categories and tags
- Links to the Tag detail page from the tag list
- List of queried events on the tag detail page
- Convert the tag list in `home_logged_in.html` into an Angular component, use Angular templates


cal/cal/urls.py
```diff
@@ -16,6 +16,9 @@
         'cal.views.complete_with_token', name='complete_with_token'),
     url(r'^accounts/profile', 'cal.views.accounts_profile', name='accounts_profile'),
 
+    url(r'categories/(?P<pk>[0-9]+)/?', 'cal.views.category_detail', name='category_detail'),
+    url(r'tags/(?P<pk>[0-9]+)/?', 'cal.views.tag_detail', name='tag_detail'),
```
jamesuejio, line 20: What does this do? ?P<pk>[0-9]?

cal/cal/templates/tag_detail.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1 class="well">Tag "{{ tag.label }}"</h1>
```
jamesuejio, line 4: Do I need to add dictionary optimization for this? Or is the tag object passed into the file, and this would be passed in through app.js where I would do the optimization.

jamesuejio (approved):
Looks good to me. I just changed the four tabs to two tabs in app.js. Also had a couple questions but besides that good 👍

cal/cal/urls.py
MaxFangX: This is regex that allows you to pass a numerical primary key into the view that is handling the request. You can look at the corresponding view in `views.py` and see that there is an extra `pk` parameter to the function.

cal/cal/templates/tag_detail.html
MaxFangX: This probably looks really confusing. This template is actually a Django template, and is using Django context variables. There is no angular code in this detail page right now. You can look at the `tag_detail` view I added to see how the `tag` context variable is getting passed into the template

MaxFangX: A few ways you can tell that it's a Django template:
- It's not in `cal/static/templates/`, just `cal/templates`. Angular templates are served as static files
- There are no verbatim tags before/after the double curly braces.

cal/cal/templates/category_detail.html
```diff
@@ -0,0 +1,6 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Category "{{ category.label }}"</h1>
```
tiffanyqi, line 4: Colon after Category? Not a merge blocker, but just wanted to point this out

tiffanyqi (commented):
There will probably be merge conflicts; let me know if you need help with anything!

cal/cal/static/js/app.js
```diff
@@ -177,29 +184,6 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
         });
       });
   };
-
-  // categories pie chart
-  $scope.options = {
```
tiffanyqi, line 182: I just realized; why did you delete this?

cal/cal/urls.py
tiffanyqi, line 20: @MaxFangX note: because we have URLs that start the same in both api and cal, cal/views.py is overriding api/views.py. I highly suggest we change the names of one or the other so that we can make post requests on categories and tags effectively.

PR #83, opened 2016-11-27
Line Visualization

jamesuejio:
Create new controller for TagDetails and add line visualization for Tags.

### DIFF
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -195,19 +195,22 @@ class ColorCategoryDetail(generics.RetrieveUpdateDestroyAPIView):
     def get_queryset(self):
         return ColorCategory.objects.filter(user=self.request.user)
 
+
 class ColorCategoryDetailEvents(generics.ListAPIView):
 
     serializer_class = GEventSerializer
 
     def get_queryset(self):
-        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return g_calendar.query()
+        calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return calendar.query()
+
 
 class ColorCategoryDetailEventWeek(APIView):
 
-    def get(self, request, *args, **kw):
-        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(g_calendar.queryWeek())
+    def get(self, request, *args, **kwargs):
+        calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return Response(calendar.get_hours_per_week())
+
 
 class TagList(generics.ListCreateAPIView):
 
@@ -261,6 +264,7 @@ class TagDetail(generics.RetrieveUpdateDestroyAPIView):
     def get_queryset(self):
         return Tag.objects.filter(user=self.request.user)
 
+
 class TagDetailEvents(generics.ListAPIView):
 
     serializer_class = GEventSerializer
@@ -269,8 +273,9 @@ class TagDetailEvents(generics.ListAPIView):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
         return tag.query()
 
+
 class TagDetailEventWeek(APIView):
 
     def get(self, request, *args, **kw):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(tag.queryWeek())
+        return Response(tag.get_hours_per_week())
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -246,3 +246,17 @@ class TagDetail(generics.RetrieveUpdateDestroyAPIView):
 
     def get_queryset(self):
         return Tag.objects.filter(user=self.request.user)
+
+class TagDetailEvents(generics.RetrieveUpdateDestroyAPIView):
+
+    serializer_class = TagSerializer
+
+    def get_serializer_context(self):
+        return {
+            'calendar_ids': self.request.query_params.get('calendar_ids'),
+            'start': self.request.query_params.get('start'),
+            'end': self.request.query_params.get('end')
+        }
+
+    def get_queryset(self):
+        return Tag.objects.filter(user=self.request.user)
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -247,16 +247,10 @@ class TagDetail(generics.RetrieveUpdateDestroyAPIView):
     def get_queryset(self):
         return Tag.objects.filter(user=self.request.user)
 
-class TagDetailEvents(generics.RetrieveUpdateDestroyAPIView):
+class TagDetailEvents(generics.ListAPIView):
 
-    serializer_class = TagSerializer
-
-    def get_serializer_context(self):
-        return {
-            'calendar_ids': self.request.query_params.get('calendar_ids'),
-            'start': self.request.query_params.get('start'),
-            'end': self.request.query_params.get('end')
-        }
+    serializer_class = GEventSerializer
 
     def get_queryset(self):
-        return Tag.objects.filter(user=self.request.user)
+        tag = Tag.objects.get(user=self.request.user, id=2) # Change this to be id of value
+        return tag.query()
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -201,15 +201,14 @@ class ColorCategoryDetailEvents(generics.ListAPIView):
     serializer_class = GEventSerializer
 
     def get_queryset(self):
-        calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return calendar.query()
+        return ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk']).query()
 
 
 class ColorCategoryDetailEventWeek(APIView):
 
     def get(self, request, *args, **kwargs):
-        calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(calendar.get_hours_per_week())
+        category = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return Response(category.get_hours_per_week())
 
 
 class TagList(generics.ListCreateAPIView):
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -8,6 +8,7 @@ from django.utils.dateparse import parse_date, parse_datetime
 from rest_framework import generics, status
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
+from rest_framework.views import APIView
 
 import pytz
 
@@ -252,5 +253,11 @@ class TagDetailEvents(generics.ListAPIView):
     serializer_class = GEventSerializer
 
     def get_queryset(self):
-        tag = Tag.objects.get(user=self.request.user, id=2) # Change this to be id of value
+        tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
         return tag.query()
+
+class TagDetailEventsWeek(APIView):
+
+    def get(self, request, *args, **kw):
+        tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return Response(tag.queryWeek())
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -195,6 +195,19 @@ class ColorCategoryDetail(generics.RetrieveUpdateDestroyAPIView):
     def get_queryset(self):
         return ColorCategory.objects.filter(user=self.request.user)
 
+class ColorCategoryDetailEvents(generics.ListAPIView):
+
+    serializer_class = GEventSerializer
+
+    def get_queryset(self):
+        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return g_calendar.query()
+
+class ColorCategoryDetailEventWeek(APIView):
+
+    def get(self, request, *args, **kw):
+        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return Response(g_calendar.queryWeek())
 
 class TagList(generics.ListCreateAPIView):
 
@@ -256,7 +269,7 @@ class TagDetailEvents(generics.ListAPIView):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
         return tag.query()
 
-class TagDetailEventsWeek(APIView):
+class TagDetailEventWeek(APIView):
 
     def get(self, request, *args, **kw):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -650,6 +650,41 @@ class Tag(models.Model, EventCollection):
 
         return events_qs.order_by('start')
 
+    def queryWeek(self, calendar_ids=None, start=None, end=None):
+        """
+        Returns a QuerySet of events matching this Tag.
+        Does not truncate at the edges.
+        """
+        calendars = self.user.profile.get_calendars_for_calendarids(calendar_ids)
+
+        keywords = self.keywords.split(',')
+        if not keywords:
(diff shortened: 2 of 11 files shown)


cal/cal/models.py
```diff
@@ -584,6 +584,40 @@ def query(self, calendar_ids=None, start=None, end=None):
         return events_qs
 
 
+    def queryWeek(self, calendar_ids=None, start=None, end=None):
+        if calendar_ids and self.calendar:
+            message = "Calendar ids can only be specified for ColorCategories without a calendar field attached."
```
tiffanyqi: what's the difference between a calendar id and a calendar field?

cal/cal/models.py
```diff
@@ -584,6 +584,40 @@ def query(self, calendar_ids=None, start=None, end=None):
...
+
+        # Union over the querysets
+        events_qs = reduce(lambda qs1, qs2: qs1 | qs2, querysets)
+
+        queryset = []
+        events = events_qs.order_by('start')
+        i = 0
+        start = events[0].start # We have to start with this outside for loop, seems sketch
+        while i < len(events):
+            end = start + timedelta(days=7)
```
tiffanyqi, line 596: this code generates the line graph data by totaling up each data point if it qualifies for a given week, right? i'm thinking we should ensure that we're within 7 days so that the events don't go past the week's threshold. is that what this code does, and the purpose of timedelta(days=7)?

cal/cal/models.py
```diff
@@ -584,6 +584,40 @@ def query(self, calendar_ids=None, start=None, end=None):
...
+            total = 0
+            while i < len(events) and (end - events[i].start).total_seconds() >= 0:
+                total += (events[i].end - events[i].start).total_seconds() / 3600
+                i += 1
```
tiffanyqi, line 600: why is i added here? why not in the larger loop?

cal/cal/models.py
```diff
@@ -650,6 +684,41 @@ def query(self, calendar_ids=None, start=None, end=None):
...
+                .exclude(all_day_event=True)
+                for keyword in keywords
+                for calendar in calendars
+                ]
+
+        # Union over the querysets
+        events_qs = reduce(lambda qs1, qs2: qs1 | qs2, querysets)
+
+        queryset = []
+        events = events_qs.order_by('start')
```
tiffanyqi: this looks like a copy and paste of the code earlier, perhaps we could generalize this code and have each class pull from that generalized code instead?

cal/cal/static/templates/tagDetails.html
```diff
@@ -0,0 +1,7 @@
+<div>
+  <nvd3 options="line" data="tagDetails"></nvd3>
```
tiffanyqi: how about you name "line" into "tagLine" instead to match "categoryLine"?

cal/cal/static/templates/categoryDetails.html
```diff
@@ -0,0 +1,7 @@
+<div>
+  <nvd3 options="categoryline" data="categoryDetails"></nvd3>
```
tiffanyqi: super picky, but shouldn't this be "categoryLine"? just to match with your other naming conventions. be sure to change the name in app.js / wherever this ends up moving!

tiffanyqi (dismissed):
Mostly a bunch of questions! Overall looks really good, looking forward to seeing this on the site :)

cal/cal/models.py
jamesuejio: I'm not sure I just copy and pasted from query( ) which has that same message.

cal/cal/models.py
jamesuejio: Yes. That's the purpose of timedelta(days=7). Basically we start at a start date, then iterate through until we hit an event that is not in the week's threshold. Then we'll set our end date to our start date and continue doing this. What's nice is if there are no events for that week, the total will be 0 and it won't go into the inner while loop.

cal/cal/models.py
jamesuejio: This is because once we hit the event that terminates the inner while loop, we want to include that event in the next iteration. For example if I had an event Monday, Tues, Sunday, Monday, we'd want to include Monday, Tues, Sunday for the first week, and Monday for the second week. If we incremented `i` after the while loop, we'd skip past the Monday.

cal/cal/models.py
jamesuejio: I just realized I can call query in my function. I'll push another change

cal/cal/models.py
jamesuejio: I thought about that but the only problem is they are in separate classes, and I'm not sure if it makes sense to have this in EventCollection or models.Model. They also call the query() method separately in both classes, so it doesn't make sense to do this. @MaxFangX any ideas?

cal/cal/static/templates/tagDetails.html
jamesuejio: Done

cal/cal/static/templates/categoryDetails.html
jamesuejio: Done

tiffanyqi:
LGTM (as far as I can tell). @MaxFangX , do you want to check again?

cal/cal/models.py
MaxFangX: calendar_id refers to the unique calendar id provided by google, and is a string. It is also the primary key of the GCalendar model. Wheneve it's just `calendar`, it's referring to the `GCalendar` objects - you can infer this from how `calendar` is set in the code.

cal/api/views.py
```diff
@@ -194,6 +195,19 @@ def get_serializer_context(self, *args, **kwargs):
     def get_queryset(self):
         return ColorCategory.objects.filter(user=self.request.user)
 
+class ColorCategoryDetailEvents(generics.ListAPIView):
+
+    serializer_class = GEventSerializer
+
+    def get_queryset(self):
+        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
```
MaxFangX: We just use `calendar` in the rest of the codebase. just a nitpick

cal/api/views.py
```diff
@@ -194,6 +195,19 @@ def get_serializer_context(self, *args, **kwargs):
...
+        return g_calendar.query()
+
+class ColorCategoryDetailEventWeek(APIView):
+
+    def get(self, request, *args, **kw):
```
MaxFangX: convention is `**kwargs`

cal/cal/models.py
```diff
@@ -584,6 +584,21 @@ def query(self, calendar_ids=None, start=None, end=None):
         return events_qs
 
 
+    def queryWeek(self, calendar_ids=None, start=None, end=None):
```
MaxFangX: `query_week` in the python standard

MaxFangX: It would be helpful to write a method comment describing what this function does, since it is not easy to tell from all the iteration code below

cal/cal/models.py
```diff
@@ -584,6 +584,21 @@ def query(self, calendar_ids=None, start=None, end=None):
...
+        queryset = []
+        events = self.query().order_by('start')
+        i = 0
+        start = events[0].start # We have to start with this outside for loop, seems sketch
+        while i < len(events):
+            end = start + timedelta(days=7)
+            total = 0
+            while i < len(events) and (end - events[i].start).total_seconds() >= 0:
+                total += (events[i].end - events[i].start).total_seconds() / 3600
```
MaxFangX, line 599: Cleaner code: instead of writing `(end - events[i].start).total_seconds()` twice, you should set it once in a variable outside the inner while loop.

cal/cal/models.py
```diff
@@ -584,6 +584,21 @@ def query(self, calendar_ids=None, start=None, end=None):
         return events_qs
 
 
+    def queryWeek(self, calendar_ids=None, start=None, end=None):
+        queryset = []
```
MaxFangX: The `query()` function is so named because it returns a Django `QuerySet` object. To prevent our code from getting really confusing, your function should either return a `QuerySet` object or not be named `query_xxx`. `get_events` looks like the closest analogy since you are returning a Python list, and `get_events` returns a set.

The reason we want to namespace like this is so that we have a reasonable expectation for when we want to do more complicated operations in the future like python set joins/intersections and/or database level joins and intersections.

My recommendation is to change the return type to a set and change this to `get_events_week` or something

cal/cal/models.py
```diff
@@ -650,6 +665,24 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         return events_qs.order_by('start')
 
+    def queryWeek(self, calendar_ids=None, start=None, end=None):
+        """
+        Returns a QuerySet of events matching this Tag.
```
MaxFangX: This does not actually return a `QuerySet`. Be careful about comments; good comment > no comment > bad comment

cal/cal/models.py
```diff
@@ -650,6 +665,24 @@ def query(self, calendar_ids=None, start=None, end=None):
...
+        start = events[0].start # We have to start with this outside for loop, seems sketch
+        while i < len(events):
+            end = start + timedelta(days=7)
+            total = 0
+            while i < len(events) and (end - events[i].start).total_seconds() >= 0:
+                total += (events[i].end - events[i].start).total_seconds() / 3600
+                i += 1
+            queryset.append((start, total))
+            start = end
+        return queryset
```
MaxFangX: All the same comments for this block of code as for the categories code above

cal/cal/templates/tag_detail.html
```diff
@@ -1,14 +1,10 @@
 {% extends "base.html" %}
 
 {% block content %}
-  <h1 class="well">Tag "{{ tag.label }}"</h1>
-  <h3 class="well">Keyword(s): "{{ tag.keywords }}"<h3>
-  <h3 class="well">Cumulative Hours: {{ tag.hours }}<h3>
-  <h3 class="well">Contained Events</h3>
```
MaxFangX, line 7: What happened to the contained events?

cal/cal/views.py
```diff
@@ -36,7 +36,6 @@ def tag_detail(request, pk):
     try:
         tag = Tag.objects.get(user=request.user, id=pk)
         context['tag'] = tag
-        context['events'] = tag.query()
```
MaxFangX, line 39: Why are we removing the list of events from the detail view?

cal/cal/static/js/app.js
```diff
@@ -108,23 +108,100 @@ analyticsApp.component('tags', {
...
+          return d3.format('.02f')(d);
+        },
+        axisLabelDistance: -10
+      },
+    },
+  };
+};
+
+analyticsApp.component('tagDetail', {
+  templateUrl: '/static/templates/tagDetails.html',
```
MaxFangX, line 180: are we doing `tagDetails` or `tagDetail`? Name of component and name of template should be consistent

cal/cal/static/js/app.js
```diff
@@ -108,23 +108,100 @@ analyticsApp.component('tags', {
   bindings: {}
 });
 
+function TagsDetailCtrl($scope, $http) {
+  var tagUrl = '/v1/tags/' + this.tagId + '/events';
+  var tagWeek = '/v1/tags/' + this.tagId + '/eventWeek';
```
MaxFangX: url convention is `eventweek` or `event-week`

cal/cal/static/js/app.js
```diff
@@ -207,7 +284,83 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
       },
     },
   };
+});
 
+function CategoriesDetailCtrl($scope, $http){
+  var categoryUrl = '/v1/colorcategories/' + this.categoryId + '/events';
+  var categoryWeek = '/v1/colorcategories/' + this.categoryId + '/eventWeek';
```
MaxFangX: same as above url convention comment

cal/cal/static/js/app.js
```diff
@@ -207,7 +284,83 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
...
+      var start = new Date(event[0]);
+      events.push({
+        x: start,
+        y: event[1]
+      });
+    }
+    $scope.categoryDetails.push({
+      values: events,      //values - represents the array of {x,y} data points
+      key: 'Category Graph', //key  - the name of the series.
+      color: '#003057',  //color - optional: choose your own line color.
```
MaxFangX: Nitpick - comments are unevenly spaced from the end of the text. Either align the start of the comments of keep consistently spacing after the commas

cal/cal/static/js/app.js
```diff
@@ -207,7 +284,83 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
...
+          return d3.format('.02f')(d);
+        },
+        axisLabelDistance: -10
+      },
+    },
+  };
+};
+
+analyticsApp.component('categoryDetail', {
+  templateUrl: '/static/templates/categoryDetails.html',
```
MaxFangX, line 358: same plurality comment as above

cal/cal/models.py
jamesuejio: Changed it to get_hours_per_week because it doesn't get the events per week, but gets the hours per week instead.

cal/cal/templates/tag_detail.html
jamesuejio: I thought we wanted to abstract that to the controllers instead of using views?

cal/cal/models.py
jamesuejio: No can do, it has to be in the while condition as the difference is changing each iteration.

cal/cal/views.py
jamesuejio: See above comment, I thought it's better style to do this in the controller instead.

cal/api/views.py
```diff
@@ -195,19 +195,22 @@ def get_serializer_context(self, *args, **kwargs):
...
 
+
 class ColorCategoryDetailEvents(generics.ListAPIView):
 
     serializer_class = GEventSerializer
 
     def get_queryset(self):
-        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return g_calendar.query()
+        calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
```
MaxFangX: `ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk']).query()`

cal/api/views.py
```diff
@@ -195,19 +195,22 @@ def get_serializer_context(self, *args, **kwargs):
...
+        return calendar.query()
+
 
 class ColorCategoryDetailEventWeek(APIView):
 
-    def get(self, request, *args, **kw):
-        g_calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(g_calendar.queryWeek())
+    def get(self, request, *args, **kwargs):
+        calendar = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
```
MaxFangX: `category`

cal/cal/models.py
```diff
@@ -583,21 +583,25 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         return events_qs
 
-
-    def queryWeek(self, calendar_ids=None, start=None, end=None):
-        queryset = []
+    def get_hours_per_week(self, calendar_ids=None, start=None, end=None):
```
MaxFangX, line 586: I see what you did here, and it makes sense. Just so we're on the same page, this is how I am understanding this change: `get_hours` is now a different type of function that returns a list of tuples of hours. In other words, it is returning time series data. So our API is now defined as follows:

- `get_events` returns a set of `GEvent`s, for python level set join/intersections
- `query` returns a Django `QuerySet`, for database level join/intersections
- `get_hours` returns time series data in the form of a list of tuples, for consumption by frontend graphs. As per our discussion today, once we generalize this to not just support weeks, I think the most appropriate name would be `get_time_series` and take in parameters something like `calendar_ids`, `start`, `end`, `time_range`.

MaxFangX (commented):
Once you implement these we are ready for +2

cal/cal/models.py
jamesuejio: Yea I like that. I'll implement get_time_series asap

MaxFangX (approved):

PR #86, opened 2016-11-27
Weekly tag stats

MaxFangX:
Implements a weekly view for Tags

- A new `CalendarRangeService` that represents the current view we're looking at - `type`, `start`, `end`, and `timeRange` (which is the key in our new `tags` dictionary). Getter and setter functions.
- `updateTags` function in tags component
- New configuration options for tags component: `isCumulative`, `displayName`, `hideZeroHours`
- Use closures instead of `bind` for the correct `this` pointer
- Rename `tags` component to `tag-list`


cal/cal/static/js/app.js
```diff
@@ -3,109 +3,171 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
+      }
+    }).success(addToList);
+  }.bind(this);
+
+  this.cancelEdit = function(tagId) {
+    var tag = this.tags[this.timeRange].find(function(tag, index, array) { return tag.id == tagId; });
+    tag.editing = false;
+  }.bind(this);
+
+  var removeFromList = function removeFromList(data) {
```
tiffanyqi: what's the difference between delete and removeFromList? why would we use one or the other right now? reason for my question: we currently delete tags and not just remove it from the list.

cal/cal/static/js/app.js
```diff
@@ -3,109 +3,171 @@ var analyticsApp = angular.module('analyticsApp', ['nvd3', 'ui.calendar']);
...
-          });
-        });
-    };
+      });
+  }
+  var initialTimeRange = CalendarRangeService.getRange()
+  this.updateTags(initialTimeRange.timeRange, initialTimeRange.start, initialTimeRange.end);
+
+  this.hideZeroHoursFilter = function (value, index, array) {
+    if (this.hideZeroHours && value.hours === 0) {
```
tiffanyqi, line 47: couldn't you instead say `return !(this.hideZeroHours && value.hours === 0)` ?

tiffanyqi (changes requested):
Finished my review! Truthfully I don't understand as much of what this does, so my comments are pretty sparse. It's probably pretty good though, be sure to test /everything/ to ensure it all still works as planned before merging!

cal/cal/static/js/app.js
MaxFangX: I think the expanded boolean is easier to read and understand, and is more in line with the semantics of the logic. Thoughts?

cal/cal/static/js/app.js
MaxFangX: Removed this variable assignment so that there isn't confusion on our controller api.

cal/api/views.py
```diff
@@ -226,10 +226,6 @@ def get_queryset(self):
         qs = Tag.objects.filter(user=self.request.user)
         start = self.request.query_params.get('start')
         end = self.request.query_params.get('end')
-        if start:
-            qs = qs.filter(start__gte=start)
-        if end:
```
jamesuejio, line 231: Will this mess up with @cindytung's filter by end date code?

MaxFangX: Nice catch. I am confused how this got into my PR

cal/cal/static/js/app.js
```diff
@@ -1,111 +1,124 @@
...
-  success(function successCallback(data) {
-    for (var i = 0; i < data.results.length; i++) {
-      var tag = data.results[i];
-      $scope.tags.push({
-        id: tag.id,
-        label: tag.label,
-        keywords: tag.keywords,
-        hours: tag.hours
-      });
+  var _this = this;
```
jamesuejio, line 11: Why do you do this?

cal/cal/static/js/app.js
```diff
@@ -1,111 +1,124 @@
...
+          id: data.id,
+          label: data.label,
+          keywords: data.keywords,
+          hours: data.hours,
+          editing: false
+        });
       });
-    });
-  };
+  }.bind(this);
```
jamesuejio, line 65: What does bind do?

cal/cal/static/js/app.js
```diff
@@ -1,111 +1,124 @@
...
 
   this.startEdit = function(tagId) {
-    var tag = $scope.tags.find(function(tag, index, array) { return tag.id == tagId; });
+    var tag = this.tags.find(function(tag, index, array) {
+      /* jshint unused:vars */
```
jamesuejio, line 69: Is there a way to disable your linter style? It sort of clutters the code base with a bunch of comments. Also isn't there a way to get rid of the unused vars? Like what is the unused var?

cal/cal/static/js/app.js
```diff
@@ -438,9 +453,10 @@ analyticsApp.controller('CalendarCtrl', function UiCalendarCtrl($scope, $http, $
 
   };
   $scope.eventRender = function(event, element, view) {
-    var location = ''
+    /* jshint unused:vars */
+    var location = '';
```
jamesuejio, line 455: See above comment.

cal/cal/static/js/services.js
```diff
@@ -0,0 +1,131 @@
+/*jslint devel: true */
+/*jslint browser: true */
+/*jslint jquery: true */
+/*global getCookie */
```
jamesuejio, line 2: Any way to disable this feature of linter? Seems like way too many excess comments. I guess it's fine but seems like there's a better fix than this.

jamesuejio (changes requested):

cal/cal/static/js/app.js
dav1dn: Essentially, `bind` "rescopes" the function to have _its_ `this` keyword set to `this` (the `this` in line 69), instead of the global scope. 

`Function.prototype.bind()` is a neater way to have a function reference a different scope without having to bind the scope to a variable.

[Further](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) [reading](https://www.smashingmagazine.com/2014/01/understanding-javascript-function-prototype-bind/)

cal/cal/static/js/services.js
tiffanyqi: +1

cal/cal/static/js/app.js
jamesuejio: Thanks makes a lot of sense.

cal/cal/static/js/app.js
```diff
@@ -1,111 +1,124 @@
...
+      tagsPromise = TagService.getTags('cumulative', null, null);
+    } else {
+      var initialTimeRange = CalendarRangeService.getRange();
+      tagsPromise = TagService.getTags(initialTimeRange.timeRange, initialTimeRange.start, initialTimeRange.end);
+    }
+    tagsPromise.then(function(tags) {
+      _this.tags = tags;
+    });
+  }.bind(this);
+  this.initialize();
```
tiffanyqi, line 43: What does 33-47 do?

cal/cal/static/js/app.js
MaxFangX: If you use `this` in a callback, it will refer to the callback instead of the controller, which is what we want. Setting a `_this` allows us to refer to the controller.

MaxFangX: `_this` is called a closure. You can read more about it [here](http://www.technofattie.com/2014/03/21/five-guidelines-for-avoiding-scope-soup-in-angular.html)

cal/cal/static/js/app.js
MaxFangX: `bind` makes the `this` included in the callback the same as the `this` where it was bound (the controller)

MaxFangX: It's an alternative to using a closure

cal/cal/static/js/app.js
MaxFangX: It gets the initial tags for the component, no matter whether it is daily/weekly/monthly/cumulative

cal/cal/static/js/services.js
MaxFangX: I think it's important to us all to use a linter; it allows us to find syntax errors a lot faster, and small things like catching when we define a function parameter that we don't actually use. If you use a linter, these will prevent your linter from eternally yelling at you.

`devel` fixes the `console` being undefined
`browser` fixes `window` being undefined
`jquery` likewise
`getCookie` is defined in our legacy javascript code - it would be really nice to angularize it into a service, but I don't think that is worth the effort for now.

How about I combine the first three into one line?

cal/cal/static/js/app.js
MaxFangX: This will prevent your own linter from yelling at you when you install it. Otherwise all us will have to configure our `.jshintrc`'s, which imo isn't worth the effort

cal/cal/static/js/app.js
MaxFangX: Same response as above!

MaxFangX: image
^What happens when you don't include these comments. Left is with comments included and right is without.

cal/cal/static/js/app.js
MaxFangX: Oh oops, didn't see the previous comments. Thanks omniscient @emesde

cal/cal/static/js/app.js
jamesuejio: Is this common style in industry? I guess I don't think I see a bunch of linter comments or what not.

cal/cal/static/js/app.js
jamesuejio: Sure fine

cal/cal/static/js/app.js
```diff
@@ -1,6 +1,4 @@
-/*jslint devel: true */
-/*jslint browser: true */
-/*jslint jquery: true */
+/*jslint devel: true, browser: true, jquery: true */
 /*global d3 */
 /*global getCookie */
 /*global moment */
```
jamesuejio: can you combine the globals into one line?

jamesuejio (changes requested):

jamesuejio (approved):

cal/cal/static/js/app.js
tiffanyqi: what's the difference between initial and not initial tags? is it just all the tags that a user wants to show up?

MaxFangX: Initial as in the current view that you're looking at. For cumulative taglist it never changes but for weekly taglist the initial would be the current list that you're looking at, then change as you press the left/right arrows on the calendar

tiffanyqi (approved):

cal/cal/static/js/app.js
MaxFangX: No!

MaxFangX: Jk yes

Issue #39, opened 2016-10-01
Add loading screen for categories and tabs


tiffanyqi:
When categories takes some time to load, create an icon that says "loading categories..." or something

tiffanyqi:
#102

PR #87, opened 2016-11-27
add preloader for categories

tiffanyqi:
Added a preloader for categories, which only shows the categories graph and respective actions when categories is fully loaded. Easy to add to other actions as well!

Temporarily used a preloader, we can change it to anything we want later on.

#39


cal/cal/static/js/app.js
```diff
@@ -111,6 +111,8 @@ analyticsApp.component('tags', {
 analyticsApp.controller('CategoriesCtrl', function($scope, $http){
   var categoryUrl = '/v1/colorcategories';
   $scope.categories = [];
+  var categories = $scope.categories;
+  categories.dataLoaded = false;
```
MaxFangX: You can just do `$scope.categories.dataLoaded = false;`

cal/cal/static/js/app.js
```diff
@@ -124,6 +126,7 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
           include: true
         });
       }
+      categories.dataLoaded = true;
```
MaxFangX: same as above comment

cal/cal/static/js/app.js
```diff
@@ -186,7 +189,7 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
   };
 
   // categories pie chart
-  $scope.options = {
+  $scope.categoryPie = {
```
MaxFangX, line 268: +1 for the name change

tiffanyqi:
@MaxFangX fixed your comments! How's it looking now?

MaxFangX (approved):
+2

PR #95, opened 2016-12-23
Excludefuture

cindytung:
Excludes future events when analyzing tags/categories. I thought doing it this way would be cleaner than changing the default end=None parameter, because I can't set end=datetime.now() since it doesn't it evaluated, and passing datetime.now() between functions seems messier.

### DIFF
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -376,6 +376,7 @@ class GEvent(Event):
 
     """
     Represents a Google event.
+    
     Reference:
     https://developers.google.com/google-apps/calendar/v3/reference/events#resource-representations
     """
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -376,7 +376,6 @@ class GEvent(Event):
 
     """
     Represents a Google event.
-    
     Reference:
     https://developers.google.com/google-apps/calendar/v3/reference/events#resource-representations
     """
@@ -582,7 +581,7 @@ class ColorCategory(models.Model, EventCollection):
         events_qs = reduce(lambda qs1, qs2: qs1 | qs2, querysets)
 
         # Exclude future events
-        events_qs = events_qs.filter(start__lt=datetime.now())
+        events_qs = events_qs.filter(start__lt=ensure_timezone_awareness(datetime.now()))
 
         return events_qs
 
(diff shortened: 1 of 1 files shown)


jamesuejio:
Did you look at line 694 `class Statistics`? Looks like it might not be used at all anywhere though so not sure. Anyways looks good to me. @MaxFangX review? There might be a more efficient way to filter it instead of union sets, then filtering it (write a function that unions AND filters), but I think for now it's fine.

jamesuejio (approved):

cal/cal/models.py
```diff
@@ -581,6 +580,9 @@ def query(self, calendar_ids=None, start=None, end=None):
         # Union over the querysets
         events_qs = reduce(lambda qs1, qs2: qs1 | qs2, querysets)
 
+        # Exclude future events
+        events_qs = events_qs.filter(start__lt=ensure_timezone_awareness(datetime.now()))
```
MaxFangX, line 584: use `utcnow()` and you won't need to use `ensure_timezone_awareness`

cal/cal/models.py
```diff
@@ -664,8 +666,8 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         if start:
             events_qs = events_qs.filter(end__gt=start)
-        if end:
```
MaxFangX, line 667: Our tag list component relies on being able to set an end time, for the daily/weekly/monthly views. If you explore to a date in the future, tags would thus be inaccurate.

cal/cal/models.py
```diff
@@ -664,8 +666,8 @@ def query(self, calendar_ids=None, start=None, end=None):
...
-            events_qs = events_qs.filter(start__lt=end)
+        if not end:
+            events_qs = events_qs.filter(start__lt=ensure_timezone_awareness(datetime.now()))
```
MaxFangX, line 670: Same `utcnow` comment as above

cal/cal/models.py
MaxFangX: We don't want to always exclude future events. There are instances where we need to set an end time in the future (see explanation in below comment)

MaxFangX (commented):
.

MaxFangX:
@jalagar to comment on your comment, this is an example of database level unions and intersections. I don't think you can make it any more efficient than it already is, because there is still only database call being made - the unions just contruct a more and more complex query.

PR #97, opened 2016-12-24
tag queries are for full words, not substrings

cindytung:
✌️

### DIFF
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -657,7 +657,7 @@ class Tag(models.Model, EventCollection):
 
         querysets = [
                 GEvent.objects
-                .filter(calendar=calendar, name__regex=r'\b(?i)'+keyword+r'\b')
+                .filter(calendar=calendar, name__regex=r'\b(?i)[#]?'+keyword+r'\b')
                 .exclude(all_day_event=True)
                 for keyword in keywords
                 for calendar in calendars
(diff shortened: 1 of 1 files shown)


cal/cal/models.py
```diff
@@ -655,7 +655,7 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         querysets = [
                 GEvent.objects
-                .filter(calendar=calendar, name__icontains=keyword)
+                .filter(calendar=calendar, name__regex=r'\b'+keyword+r'\b')
```
MaxFangX: omfg I never knew you could do this. You're amazing

MaxFangX (approved):
Can you confirm that this is still case-insensitive? Otherwise feel free to merge (:

cal/cal/models.py
cindytung: ahh you're right, I didn't make it case insensitive. still new to regex :)

cal/cal/models.py
```diff
@@ -655,7 +655,7 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         querysets = [
                 GEvent.objects
-                .filter(calendar=calendar, name__regex=r'\b'+keyword+r'\b')
+                .filter(calendar=calendar, name__regex=r'\b(?i)'+keyword+r'\b')
```
jamesuejio: Impressive. I think anything that deals with regex is magic

jamesuejio (approved):

PR #98, opened 2016-12-25
fix timezone issues

cindytung:
💯

### DIFF
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -583,7 +583,7 @@ class ColorCategory(models.Model, EventCollection):
         if end:
             events_qs = events_qs.filter(start__lt=end)
         else:
-            events_qs = events_qs.filter(start__lt=datetime.utcnow())
+            events_qs = events_qs.filter(start__lt=datetime.now(pytz.utc))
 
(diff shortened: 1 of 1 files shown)


cal/cal/models.py
```diff
@@ -583,7 +583,7 @@ def query(self, calendar_ids=None, start=None, end=None):
         if end:
             events_qs = events_qs.filter(start__lt=end)
         else:
-            events_qs = events_qs.filter(start__lt=datetime.utcnow())
+            events_qs = events_qs.filter(start__lt=datetime.now(pytz.utc))
```
jamesuejio, line 587: Doesn't work you need to `import pytz`

jamesuejio (changes requested):

cal/cal/models.py
cindytung: wth how does this not throw errors

jamesuejio: yay Python. naw idk I figured it out cuz the graph and events didn't show up, then I clicked something and it showed an error page.

jamesuejio (approved):

PR #101, opened 2016-12-28
More Line graph visualization

jamesuejio:


cal/cal/helpers.py
```diff
@@ -28,6 +31,43 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
...
+    if timezone:
+        time = time.astimezone(timezone)
+    else:
+        time = time.astimezone(timezone_util.utc)
+
+    return time
+
+
+def local_to_UTC(time, timezone):
+    return time.astimezone(pytz.utc)
```
MaxFangX: I don't think we should be using a helper function here, because it doesn't have any additional functionality. Imo we should use the datetime libraries directly because they all have pretty clean APIs and it's universally understood by any python developers who have worked with time.

cal/api/serializers.py
```diff
@@ -82,3 +82,9 @@ def get_hours(self, obj):
         start = self.context.get('start')
         end = self.context.get('end')
         return obj.hours(calendar_ids=calendar_ids, start=start, end=end)
+
+
+class ColorCategoryList(serializers.ModelSerializer):
```
MaxFangX: This should be named something with `Serializer` at the end. `List` always refers to a view, and thus causes confusion. Maybe `ColorCategoryTimeSeriesSerializer`?

Was there something like this in the previous PR?

cal/api/views.py
```diff
@@ -117,37 +114,8 @@ def get_queryset(self):
...
-                time = time.astimezone(timezone)
-            else:
-                time = time.astimezone(timezone_util.utc)
-
-            return time
-
-        start = handle_time_string(start_str)
-        end = handle_time_string(end_str)
+        start = handle_time_string(start_str, timezone_str)
+        end = handle_time_string(end_str, timezone_str)
```
MaxFangX, line 118: yesss this is so much cleaner

cal/cal/helpers.py
```diff
@@ -1,7 +1,10 @@
 from django.core.serializers.json import DjangoJSONEncoder
 from django.http import HttpResponse
 from django.utils import timezone
+from django.utils.dateparse import parse_date, parse_datetime
+from django.utils import timezone as timezone_util
 from datetime import timedelta
+from datetime import datetime
```
MaxFangX: - Duplicate import of timezone
- `datetime, timedelta`

cal/api/views.py
```diff
@@ -201,14 +169,17 @@ class ColorCategoryDetailEvents(generics.ListAPIView):
     serializer_class = GEventSerializer
 
     def get_queryset(self):
-        return ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk']).query()
+        category = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return category.query()
```
MaxFangX: I don't think there is a good reason to change this from what it was before. The one liner is cleaner than setting a temporary variable

cal/cal/models.py
```diff
@@ -583,20 +584,38 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         return events_qs
 
-    def get_hours_per_week(self, calendar_ids=None, start=None, end=None):
+    def get_hours_per_week(self, timezone, calendar_ids=None, start=None, end=None):
```
MaxFangX: timezone should default to UTC.

cal/api/views.py
```diff
@@ -201,14 +169,17 @@ class ColorCategoryDetailEvents(generics.ListAPIView):
...
 
 
 class ColorCategoryDetailEventWeek(APIView):
 
+    serializer_class = ColorCategoryList
+
     def get(self, request, *args, **kwargs):
         category = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(category.get_hours_per_week())
+        return Response(category.get_hours_per_week(self.request.query_params.get('timezone')))
```
MaxFangX: Be wary of the future when we want to be able to get the hours per time period for events within a certain time range. For example, if we had a more detailed "day" view in our chart and scrolled back in time, and now need to hit our api with a start and end time. This would mean we have to specify the start and end time, which may or may not have a timezone attached to it. Given that the timezone itself may or may not be specified, how do we managed the relationship between the two?

`handle_time_string` already covers most of what I just said, I just wanted to give you a heads up.

cal/cal/helpers.py
```diff
@@ -28,6 +31,43 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
     return dt
 
 
+def handle_time_string(time_str, timezone_str):
+    """
+    Converts UTC 'time_str' to timezone represented by 'timezone_str'
```
MaxFangX: This description need to be improved. If you look closely at the logic, the timezone is first ignored if a timezone could be parsed from the time string. Then the timezone is applied to the time so that it is both timezone aware and has a timezone offset. It is not always converted, since we don't want to convert datetimes that are already timezone aware.

cal/cal/helpers.py
```diff
@@ -406,5 +446,5 @@ def old_insert(self, timenode):
                 self.next.insert(timenode)
             elif self.prev:
                 self.prev.insert(timenode)
-            
+
```
MaxFangX, line 510: Thanks for these

cal/cal/models.py
```diff
@@ -583,20 +584,38 @@ def query(self, calendar_ids=None, start=None, end=None):
...
         """
         Returns a list of week-hour tuples corresponding to the events in this ColorCategory.
         Each week starts at the start time.
         """
         week_hours = []
         events = self.query().order_by('start')
         i = 0
-        start = events[0].start
+        # Convert start to local time
+        start = handle_time_string(str(events[0].start), timezone)
```
MaxFangX: Events saved in our database are already ensured to be timezone aware (see `GEvent.save`) so we shouldn't be using `handle_time_string` because that contains a lot of extra logic designed for any kind of input. Instead you can just something like `start.astimezone(pytz.timezone(timezone_str))`.

cal/cal/models.py
```diff
@@ -583,20 +584,38 @@ def query(self, calendar_ids=None, start=None, end=None):
...
+        # Change start date to be Monday beginning of week
+        while start.weekday() != 0:
+            start = start - relativedelta(days=1)
```
MaxFangX: You should use datetime.timedelta because we're already using that everywhere else in our codebase. I don't think there is any place where we need to use dateutil because the django time utilities + python datetime + pytz are more than enough for everything we need, and we should try to stay consistent with our libraries if we can.

cal/cal/models.py
MaxFangX: You're not actually using the `start` that was passed into this function as a parameter; you're overriding it and going by the first event that is returned in the query, though you're probably already aware of this. This is best for cumulative queries but breaks if we want to zoom and look at subsets of the overall query (ex. I want to look at only April from this year)

MaxFangX: You are not using `calendar_ids` either. There are other example implementations

MaxFangX (commented):
WIP review

cal/cal/helpers.py
jamesuejio: Should I then import pytz in models.py?

cal/api/serializers.py
jamesuejio: No good catch.

cal/api/views.py
jamesuejio: Fine

cal/api/views.py
jamesuejio: Wait I'm a bit confused, so what do I need to change here?

cal/cal/models.py
jamesuejio: so `timezone='UTC'`? just want to make sure

cal/cal/models.py
jamesuejio: Sure I'll change it for this but I need it for `end = start + relativedelta(days=7)` because I'm going to be generalizing it for months and relativedelta makes sure that adding to months will always start on the same day.

cal/cal/helpers.py
```diff
@@ -28,6 +31,90 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
...
+        # To indicate do nothing if Daily is passed in
+        pass
+    if time_range == 'Weekly':
+        # Change start date to be Monday beginning of week
+        while start.weekday() != 0:
+            start = start - timedelta(days=1)
+    if time_range == 'Monthly':
+        # Change start date to beginning of month
+        while start.day != 1:
+            start = start - timedelta(days=1)
```
jamesuejio, line 90: @cindytung Here's the updated generalized function that can take in 'Daily', 'Weekly', "Monthly' and will do everything accordingly :) all you need to do is call this function with different arguments in your next PR.

cal/cal/helpers.py
jamesuejio: Ok adding.

cal/cal/helpers.py
cindytung: WOoHOO ty jamesiez this is awesome

cal/cal/helpers.py
MaxFangX: Yes!

cal/api/views.py
MaxFangX: Nothing! I was just making comments

cal/cal/models.py
MaxFangX: I believe that will work, but I'd test it just to be sure.

MaxFangX: There's probably a way to get the string representation of a timezone using pytz

cal/cal/models.py
MaxFangX: Ahh nice, I didn't know that.

cal/api/views.py
```diff
@@ -206,9 +174,11 @@ def get_queryset(self):
 
 class ColorCategoryDetailEventWeek(APIView):
```
MaxFangX: This still says week

cal/api/views.py
```diff
@@ -275,6 +245,8 @@ def get_queryset(self):
 
 class TagDetailEventWeek(APIView):
```
MaxFangX: It still says week

cal/cal/helpers.py
```diff
@@ -1,7 +1,10 @@
 from django.core.serializers.json import DjangoJSONEncoder
 from django.http import HttpResponse
 from django.utils import timezone
```
MaxFangX: The duplicate timezone import is still here

cal/cal/helpers.py
```diff
@@ -28,6 +31,93 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
     return dt
 
 
+def handle_time_string(time_str, timezone_str):
+    """
+    If timezone_str exists get the corresponding timezone. Parses time_str
+    and if there is no time, create an event at 0th hour.
+    Makes timezone aware if unaware then converts to corresponding timezone.
```
MaxFangX, line 38: This is fantastic, thank you

cal/cal/helpers.py
```diff
@@ -28,6 +31,93 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
...
+
+    if timezone:
+        time = time.astimezone(timezone)
+    else:
+        time = time.astimezone(timezone_util.utc)
+
+    return time
+
+
+def get_time_series(model, timezone='UTC', time_range='Weekly', start=None, end=None):
```
MaxFangX: `time_range` is a little bit confusing because we have `start` and `end` defining a range - what about `time_increment`? `time_period`?

MaxFangX: `time_step`? FullCalendar uses `type` to denote `day`, `month`, `agendaWeek` I trust your judgment

cal/cal/helpers.py
```diff
@@ -28,6 +31,93 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
...
+    """
+    Returns a list of week-hour tuples corresponding to the events in the `model`. Takes in
+    timezone in order to accurately aggregate events. Includes time_range input either `Daily`,
+    `Weekly`, or `Monthly` which will aggregate accordingly. Splices events that overlap times.
```
MaxFangX: Nitpick - when we use strings as a basis of comparison and logic, the standard is to always use all lowercase, because it simplifies things.

cal/cal/helpers.py
```diff
@@ -28,6 +31,93 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
...
+    """
+    week_hours = []
+    events = model.query().order_by('start')
+    i = 0
+    # Convert start to local time
+    start = handle_time_string(str(events[0].start), timezone)
```
MaxFangX: Why didn't you do the `astimezone` thing?

MaxFangX: Reminder that you still need to implement the `start` and `end` parameters, as well as `calendar_ids`

cal/cal/helpers.py
```diff
@@ -28,6 +31,93 @@ def ensure_timezone_awareness(dt, optional_timezone=None):
...
+    rollover = 0
+    while i < len(events):
+        if time_range == 'Daily':
+            end = start + relativedelta(days=1)
+        if time_range == 'Weekly':
+            end = start + relativedelta(days=7)
+        if time_range == 'Monthly':
+            end = start + relativedelta(months=1)
+        # Deal with daylight savings time
+        if handle_time_string(str(end), timezone).hour != 0:
```
MaxFangX: This is another example of overapplication of `handle_time_string`. You should find a way to do it using the standard date/time libraries.

cal/cal/models.py
```diff
@@ -669,23 +657,11 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         return events_qs.order_by('start')
 
-    def get_hours_per_week(self, calendar_ids=None, start=None, end=None):
+    def get_time_series(self, timezone='UTC', time_range='Weekly', start=None, end=None):
```
MaxFangX: Same time_range comment as above

cal/cal/static/js/app.js
```diff
@@ -111,29 +111,41 @@ analyticsApp.component('tags', {
...
   $scope.tagEvents = [];
+  $scope.tagHours = this.tagHours;
 
   $http({method: 'GET', url: tagUrl + '.json' }).
   success(function successCallback(data) {
     for (var i = 0; i < data.results.length; i++) {
       var event = data.results[i]
       $scope.tagEvents.push({
-        start: event.start,
+        start: (new Date(event.start)).toString(),
```
MaxFangX, line 131: Does `toISOString` still work? If so, I think we should use `toISOString` because it's the less 'naive' version. In javascript datetimes at least, ISOString gives you a timezone aware string, so I'm wondering if that same principle can be applied here.

cal/cal/static/js/app.js
```diff
@@ -111,29 +111,41 @@ analyticsApp.component('tags', {
...
-
-  $http({method: 'GET', url: eventweek + '.json' }).
+  $http({method: 'GET',
+         url: eventweek + '.json',
+         params: {
+           timezone: query_timezone,
+         }
+       }).
   success(function successCallback(data) {
+    $scope.tagHours = $scope.tagHours/data.length;
```
MaxFangX: Spaces around the `/` operator please~

cal/cal/static/js/app.js
```diff
@@ -111,29 +111,41 @@ analyticsApp.component('tags', {
...
       $scope.tagEvents.push({
-        start: event.start,
+        start: (new Date(event.start)).toString(),
         name: event.name,
       });
     }
   });
-
-  $http({method: 'GET', url: eventweek + '.json' }).
+  $http({method: 'GET',
```
MaxFangX: Nitpick: If we don't have all of the key-value pairs inline, the common style is to put all the key-value pairs starting on the next line.

cal/cal/static/js/app.js
```diff
@@ -111,29 +111,41 @@ analyticsApp.component('tags', {
 function TagsDetailCtrl($scope, $http) {
   var tagUrl = '/v1/tags/' + this.tagId + '/events';
   var eventweek = '/v1/tags/' + this.tagId + '/eventWeek';
+  var query_timezone = moment.tz.guess();
   $scope.tagDetails = [];
   $scope.tagEvents = [];
+  $scope.tagHours = this.tagHours;
```
MaxFangX, line 124: For these three variables, change it to `this.tagDetails` etc, binding to the controller. Angular is moving away from the usage of scope for a variety of reasons such as it makes inheritance confusing, it pollutes the global namespace. After you've bound to your controller, in your html you can refer to it as `$ctrl.tagDetails`, or `TagsDetailCtrl.tagDetails` if you haven't used the `controllerAs` configuration.

[Link](https://toddmotto.com/no-scope-soup-bind-to-controller-angularjs/)
[Link 2](http://www.technofattie.com/2014/03/21/five-guidelines-for-avoiding-scope-soup-in-angular.html)

cal/cal/static/js/app.js
```diff
@@ -111,29 +111,41 @@ analyticsApp.component('tags', {
...
+         url: eventweek + '.json',
+         params: {
+           timezone: query_timezone,
+         }
+       }).
   success(function successCallback(data) {
+    $scope.tagHours = $scope.tagHours/data.length;
     var events = [];
+    var max_hour = 0;
     for (var i = 0; i < data.length; i++) {
```
MaxFangX, line 147: Some explanatory comments around this area would be particularly helpful, since it seems to be somewhat math/logic heavy.

cal/cal/static/js/app.js
```diff
@@ -142,46 +154,47 @@ function TagsDetailCtrl($scope, $http) {
...
-        axisLabelDistance: -10
+        x: function(d){ return d.x; },
+        y: function(d){ return d.y; },
+        useInteractiveGuideline: true,
+        xScale: d3.time.scale(),
+        xAxis: {
+          axisLabel: 'Date',
+          tickFormat: function(d) {
+                          return d3.time.format('%m/%d/%y')(d)
+                      }
```
MaxFangX: Lines 175 and 176 are indented incorrectly

cal/cal/static/js/app.js
MaxFangX: Btw, `$scope.tagHours` is redundant because it already exists as a binding to your controller (as configured in your component)

cal/cal/static/js/app.js
```diff
@@ -289,29 +302,42 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
...
   $scope.categoryEvents = [];
+  $scope.categoryHours = this.categoryHours;
 
   $http({method: 'GET', url: categoryUrl + '.json' }).
   success(function successCallback(data) {
     for (var i = 0; i < data.results.length; i++) {
       var event = data.results[i]
       $scope.categoryEvents.push({
-        start: event.start,
+        start: (new Date(event.start)).toString(),
```
MaxFangX, line 347: same `toISOString()` comment as above

cal/cal/static/js/app.js
```diff
@@ -289,29 +302,42 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
...
         name: event.name,
       });
     }
   });
 
-  $http({method: 'GET', url: eventweek + '.json' }).
+  $http({method: 'GET',
```
MaxFangX: Same indentation comment as above

cal/cal/static/js/app.js
```diff
@@ -289,29 +302,42 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
...
+         url: eventweek + '.json',
+         params: {
+           timezone: query_timezone,
+         }
+       }).
   success(function successCallback(data) {
+    $scope.categoryHours = $scope.categoryHours/data.length;
```
MaxFangX: same `/`

cal/cal/static/js/app.js
```diff
@@ -289,29 +302,42 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
...
     var events = [];
+    var max_hour = 0;
     for (var i = 0; i < data.length; i++) {
```
MaxFangX, line 364: same make comments here math logic heavy

cal/cal/static/js/app.js
```diff
@@ -320,46 +346,48 @@ function CategoriesDetailCtrl($scope, $http){
...
+        },
+        x: function(d){ return d.x; },
+        y: function(d){ return d.y; },
+        useInteractiveGuideline: true,
+        xScale: d3.time.scale(),
+        xAxis: {
+          axisLabel: 'Date',
+          tickFormat: function(d) {
+                          return d3.time.format('%m/%d/%y')(d)
+                      }
```
MaxFangX: same wrong indentation

cal/cal/static/js/app.js
```diff
@@ -320,46 +346,48 @@ function CategoriesDetailCtrl($scope, $http){
...
-        tickFormat: function(d) {
-                        return d3.time.format('%m/%d/%y')(d)
-                    }
-      },
-      yAxis: {
-        axisLabel: 'Hours',
-        tickFormat: function(d){
-          return d3.format('.02f')(d);
+    // line graph
+    $scope.categoryLine = {
```
MaxFangX, line 383: Bind to controller

cal/cal/static/js/app.js
```diff
@@ -289,29 +302,42 @@ analyticsApp.controller('CategoriesCtrl', function($scope, $http){
 function CategoriesDetailCtrl($scope, $http){
   var categoryUrl = '/v1/colorcategories/' + this.categoryId + '/events';
   var eventweek = '/v1/colorcategories/' + this.categoryId + '/eventWeek';
+  var query_timezone = moment.tz.guess();
   $scope.categoryDetails = [];
   $scope.categoryEvents = [];
+  $scope.categoryHours = this.categoryHours;
```
MaxFangX, line 340: same bind to controller, not scope

MaxFangX: Same `categoryHours` redundant

cal/api/views.py
jamesuejio: Changed

cal/api/views.py
jamesuejio: Changed

cal/cal/helpers.py
jamesuejio: Alright I used timezone_util instead just because it's clearer

cal/cal/helpers.py
jamesuejio: Using time_step

cal/cal/helpers.py
jamesuejio: fine i'll make it all lower. makes sense.

cal/cal/helpers.py
jamesuejio: I added back the calender_ids parameter but will implement next PR.

jamesuejio: And yea I did it with astimezone missed the old comment

cal/cal/models.py
jamesuejio: yea changed timezone logic.

cal/cal/helpers.py
jamesuejio: Changed

cal/cal/models.py
jamesuejio: Done

cal/cal/static/js/app.js
jamesuejio: Changing it to `this.tagDetails = []` and `this.tagDetails.push` gives an angular error saying Cannot read property 'push' of undefined... I changed it in the html to `TagsDetailCtrl.tagDetails` as well. am I doing it wrong, do I have to do more?

cal/cal/static/js/app.js
jamesuejio: No it doesn't. I tried event.start.toISOString, tried putting it after new Date, tried some other things nothing works like this does.

cal/cal/static/js/app.js
jamesuejio: Done

cal/cal/static/js/app.js
jamesuejio: Ya

cal/cal/static/js/app.js
jamesuejio: Ya

cal/cal/static/js/app.js
jamesuejio: okie

cal/cal/static/js/app.js
jamesuejio: See above comment and question

cal/cal/static/js/app.js
jamesuejio: doesn't work

cal/cal/static/js/app.js
jamesuejio: Will wait to do this. Actually i think Cindy will do this in her PR because she refactored a bunch of line graph logic and I think she binded to controller. @cindytung JS god

cal/cal/static/js/app.js
jamesuejio: Fixed

cal/cal/static/js/app.js
jamesuejio: Ya / / / / /

cal/cal/static/js/app.js
jamesuejio: yup

cal/cal/static/js/app.js
MaxFangX: Notice that in the configuration of your component you have `controllerAs: '$ctrl'`, which means that you should refer to your variable using `$ctrl.tagDetails`. I believe that should solve the issue

cal/cal/static/js/app.js
MaxFangX: Ok thanks for checking

cal/cal/static/js/app.js
MaxFangX: / / / /

cal/cal/static/js/app.js
MaxFangX: Okay, as long as @cindytung is aware of it!

MaxFangX (approved):
Given @cindytung is doing the bindings to controller part, LGTM (:

cal/cal/static/js/app.js
jamesuejio: The problem isn't referring it in the html file, but changing the instances in app.js to be this.tagDetails. It gives an error. I'm just going to merge so Cindy can work on it, and I'll work on scope soup in my next PR

PR #102, opened 2016-12-28
add preloader for tags, tagDetail, and categoryDetail

tiffanyqi:
Preloader for tags, tagDetail, and categoryDetail. Also accounts for tag or category deletion.

There's an infinite spinner for tags-weekly, but I'm sure this will be fixed once we resolve "failed to get tags"


MaxFangX (approved):
lgtm

MaxFangX:
+2 for slack notif

MaxFangX:
Typically when you're merging, you don't want to change the name of the commit because you want it to be clear that it was a merge commit, even if you had to resolve a bunch of merge conflicts. You can see that it's a merge commit on Github but it's not so obvious in `git log`

tiffanyqi:
@MaxFangX ok! i'll be better about that :)

PR #105, opened 2016-12-29
[WIP] Auto full sync when a user logs in

### LINKED ISSUE #11: Automatic syncing for users/superusers
Superusers currently click on "full sync all" in order to see their week(s), and I'm not sure how normal users are able to do this.

Idea is to create automatic syncing for one week, depending on what week the user is looking at. Google Calendar seems to do this well.

As for calculating hours, this could be done by loading and calculating hours in accordance with their query (four weeks in the past, etc) on the back, rather than needing to sync _all_ events.

cindytung:
Fixes #11

### DIFF
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -28,6 +28,7 @@ class Profile(models.Model):
     main_calendar = models.ForeignKey("GCalendar", null=True)
     authed = models.BooleanField(default=False, help_text="If the user's oauth credentials are currently valid")
     analysis_start = models.DateTimeField(null=True, blank=True, help_text="When the analysis of the user's calendar will start")
+    first_time_authed = models.BooleanField(default=False, help_text="If it's the first time the user's oauth credentials are authenticated")
 
     created_at = models.DateTimeField(auto_now_add=True)
     modified_at = models.DateTimeField(auto_now=True)
@@ -127,6 +128,9 @@ class Profile(models.Model):
             if qs.count() > 0:
                 create_category_if_nonexistent(color_index="1", gcalendar=calendar)
 
+    def resetFirstTimeAuth(self):
+        self.first_time_authed = False
+        self.save()
 
 class GCalendar(models.Model):
 
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -27,7 +27,7 @@ class Profile(models.Model):
     main_calendar = models.ForeignKey("GCalendar", null=True)
     authed = models.BooleanField(default=False, help_text="If the user's oauth credentials are currently valid")
     analysis_start = models.DateTimeField(null=True, blank=True, help_text="When the analysis of the user's calendar will start")
-    first_time_authed = models.BooleanField(default=False, help_text="If it's the first time the user's oauth credentials are authenticated")
+    # first_time_authed = models.BooleanField(default=False, help_text="If it's the first time the user's oauth credentials are authenticated")
 
     created_at = models.DateTimeField(auto_now_add=True)
     modified_at = models.DateTimeField(auto_now=True)
@@ -127,9 +127,9 @@ class Profile(models.Model):
             if qs.count() > 0:
                 create_category_if_nonexistent(color_index="1", gcalendar=calendar)
 
-    def resetFirstTimeAuth(self):
-        self.first_time_authed = False
-        self.save()
+    # def resetFirstTimeAuth(self):
+    #     self.first_time_authed = False
+    #     self.save()
 
 class GCalendar(models.Model):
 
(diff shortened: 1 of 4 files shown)


cal/cal/templates/home_logged_in.html
```diff
@@ -26,6 +26,13 @@
         }
       });
     });
+    function runFullSync() {
+      if ("{{request.user.profile.first_time_authed}}" == "True") {
```
MaxFangX: You'll want a triple equals here (`===`). See [link](https://stackoverflow.com/questions/523643/difference-between-and-in-javascript)

cal/cal/templates/home_logged_in.html
```diff
@@ -26,6 +26,13 @@
...
+        "{{request.user.profile.resetFirstTimeAuth}}";
+        window.location = "/v1/sync?full_sync=true&sync_all=true";
+      };
+    };
+    window.onload = runFullSync;
```
MaxFangX: Since you don't always run full sync on page load, the name of this function might be confusing. Maybe something like "syncIfFirstTimeAuth" or something? I trust your judgment.

Edit: This might not be relevant anymore after you angularize this code and understand the implications of inc sync

cal/cal/templates/home_logged_in.html
```diff
@@ -26,6 +26,13 @@
         }
       });
     });
+    function runFullSync() {
```
MaxFangX: Do you think you could this stuff in Angular instead? All of the other javascript in `home_logged_in.html` is copy pasted stuff from auth documentation, and we're trying to avoid putting arbitrary logic in the html (and usually when we do, it's in `calendaranalytics.js` - that's where the legacy javascript is). Some things you might want to check out:

- `LoggedInController` which is currently unused but would be perfectly suited for Angularizing this code
- `ng-init` for injecting data like you have using Django templating
- `$location` is the Angular way of changing the window location. We don't want to use pure javascript in this case. [Docs](https://docs.angularjs.org/guide/$location)

Thanks! Excited for this :D

cal/cal/templates/home_logged_in.html
```diff
@@ -26,6 +26,13 @@
...
+      if ("{{request.user.profile.first_time_authed}}" == "True") {
+        "{{request.user.profile.resetFirstTimeAuth}}";
+        window.location = "/v1/sync?full_sync=true&sync_all=true";
```
MaxFangX: I don't think you want `full_sync=true` here. If inc sync works then that is way faster than full sync. Recall that if you call inc sync and it doesn't work, then full sync will run anyway. We only want to full sync when we want a clean reset of every event.

cal/cal/templates/home_logged_in.html
```diff
@@ -26,6 +26,13 @@
         }
       });
     });
+    function runFullSync() {
+      if ("{{request.user.profile.first_time_authed}}" == "True") {
+        "{{request.user.profile.resetFirstTimeAuth}}";
```
MaxFangX: I think you are just referencing the function here instead of calling it. I might be wrong though since I haven't verified it myself.

cal/cal/models.py
```diff
@@ -28,6 +28,7 @@ class Profile(models.Model):
     main_calendar = models.ForeignKey("GCalendar", null=True)
     authed = models.BooleanField(default=False, help_text="If the user's oauth credentials are currently valid")
     analysis_start = models.DateTimeField(null=True, blank=True, help_text="When the analysis of the user's calendar will start")
+    first_time_authed = models.BooleanField(default=False, help_text="If it's the first time the user's oauth credentials are authenticated")
```
MaxFangX: You should commit your migrations too. Other people on the project should only need to run `python manage.py migrate`.

cal/cal/views.py
```diff
@@ -126,6 +126,7 @@ def google_auth(request):
         storage.put(credential)
         profile, _ = Profile.get_or_create(user=request.user)
         profile.authed = True
+        profile.first_time_authed = True
```
MaxFangX, line 129: I actually don't think you need the `first_time_authed` field. If you run incremental sync every time the user logs in it will already handle all the behavior you need. In the case where it's the user's first time logging in ever, inc sync will try to get a sync token, fail, and then run a full sync. In the normal case, when it's not the user's first time logging in, inc sync will work fine (after we fix it of course).

cal/cal/models.py
MaxFangX: Also be wary of asking other people to migrate before your branch has been pushed into master. If you push your migrations (the generated code) their local environments will see that the migration has not been run and attempt to run it, but find that the field `first_time_authed` already exists.

Why I say "pushed into master" is because that's when all the migrations have been assumed to be safe and finalized, so our production should run the migration, as well as the other devs.

Anyone who has run this migration already will likely need to enter their mysql shell and drop the column created by this field. Dw about it though, we can figure it out if it comes up

MaxFangX (changes requested):

cal/cal/views.py
cindytung: Ohhh ok I was under the impression that my task was to run a sync ONLY when after a first time log in, and do nothing for all other logins. Thanks I think the fix should be pretty simple

PR #106, opened 2016-12-31
Added daily/weekly/monthly timeline buttons

cindytung:
:neckbeard:

### DIFF
--- a/cal/api/urls.py
+++ b/cal/api/urls.py
@@ -15,12 +15,12 @@ urlpatterns = patterns(
     url(r'^v1/colorcategories/$', views.ColorCategoryList.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/?$', views.ColorCategoryDetail.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/events/?$', views.ColorCategoryDetailEvents.as_view()),
-    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/eventWeek/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
+    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/event(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
     url(r'^v1/stats/$', views.StatisticList.as_view()),
     url(r'^v1/tags/$', views.TagList.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/?$', views.TagDetail.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/events/?$', views.TagDetailEvents.as_view()),
-    url(r'^v1/tags/(?P<pk>[0-9]+)/eventWeek/?$', views.TagDetailEventTimeSeries.as_view()),
+    url(r'^v1/tags/(?P<pk>[0-9]+)/event(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
 
 )
 
--- a/cal/api/urls.py
+++ b/cal/api/urls.py
@@ -15,12 +15,12 @@ urlpatterns = patterns(
     url(r'^v1/colorcategories/$', views.ColorCategoryList.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/?$', views.ColorCategoryDetail.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/events/?$', views.ColorCategoryDetailEvents.as_view()),
-    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/event(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
+    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/event/(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
     url(r'^v1/stats/$', views.StatisticList.as_view()),
     url(r'^v1/tags/$', views.TagList.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/?$', views.TagDetail.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/events/?$', views.TagDetailEvents.as_view()),
-    url(r'^v1/tags/(?P<pk>[0-9]+)/event(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
+    url(r'^v1/tags/(?P<pk>[0-9]+)/event/(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
 
 )
 
--- a/cal/api/urls.py
+++ b/cal/api/urls.py
@@ -15,12 +15,12 @@ urlpatterns = patterns(
     url(r'^v1/colorcategories/$', views.ColorCategoryList.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/?$', views.ColorCategoryDetail.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/events/?$', views.ColorCategoryDetailEvents.as_view()),
-    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/event/(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
+    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/timeseries/(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
     url(r'^v1/stats/$', views.StatisticList.as_view()),
     url(r'^v1/tags/$', views.TagList.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/?$', views.TagDetail.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/events/?$', views.TagDetailEvents.as_view()),
-    url(r'^v1/tags/(?P<pk>[0-9]+)/event/(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
+    url(r'^v1/tags/(?P<pk>[0-9]+)/timeseries/(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
 
 )
 
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -178,7 +178,7 @@ class ColorCategoryDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kwargs):
         category = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(category.get_time_series(self.request.query_params.get('timezone')))
+        return Response(category.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
 
 
 class TagList(generics.ListCreateAPIView):
@@ -242,4 +242,4 @@ class TagDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kw):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(tag.get_time_series(self.request.query_params.get('timezone')))
+        return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -178,7 +178,7 @@ class ColorCategoryDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kwargs):
         category = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(category.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step'].lower()))
+        return Response(category.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
 
 
 class TagList(generics.ListCreateAPIView):
@@ -242,4 +242,4 @@ class TagDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kw):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step'].lower()))
+        return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
--- a/cal/api/views.py
+++ b/cal/api/views.py
@@ -178,7 +178,7 @@ class ColorCategoryDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kwargs):
         category = ColorCategory.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(category.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
+        return Response(category.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step'].lower()))
 
 
 class TagList(generics.ListCreateAPIView):
@@ -242,4 +242,4 @@ class TagDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kw):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
+        return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step'].lower()))
--- a/cal/cal/helpers.py
+++ b/cal/cal/helpers.py
@@ -66,7 +66,7 @@ def handle_time_string(time_str, timezone_str):
     return time
 
 
-def get_time_series(model, timezone='UTC', time_step='weekly', calendar_ids=None, start=None, end=None):
+def get_time_series(model, timezone='UTC', time_step='Week', calendar_ids=None, start=None, end=None):
     """
     Returns a list of week-hour tuples corresponding to the events in the `model`. Takes in
     timezone in order to accurately aggregate events. Includes time_step input either `daily`,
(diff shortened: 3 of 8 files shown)


cal/cal/helpers.py
```diff
@@ -66,7 +66,7 @@ def handle_time_string(time_str, timezone_str):
     return time
 
 
-def get_time_series(model, timezone='UTC', time_step='weekly', calendar_ids=None, start=None, end=None):
+def get_time_series(model, timezone='UTC', time_step='Week', calendar_ids=None, start=None, end=None):
```
jamesuejio: Lower case for better style (just repeating what Max told me)

cal/cal/helpers.py
```diff
@@ -77,14 +77,14 @@ def get_time_series(model, timezone='UTC', time_step='weekly', calendar_ids=None
     i = 0
     # Convert start to local time
     start = events[0].start.astimezone(pytz.timezone(timezone))
-    if time_step == 'daily':
+    if time_step == 'Day':
```
jamesuejio: change all this to lower case :)

cal/cal/static/js/app.js
```diff
@@ -180,20 +179,83 @@ function TagsDetailCtrl($scope, $http) {
         xAxis: {
           axisLabel: 'Date',
           tickFormat: function(d) {
-            return d3.time.format('%m/%d/%y')(d)
-          }
+                          return d3.time.format('%m/%d/%y')(d)
+                      }
```
jamesuejio: change indentation back

cal/cal/static/js/app.js
```diff
@@ -180,20 +179,83 @@ function TagsDetailCtrl($scope, $http) {
...
         yAxis: {
           axisLabel: 'Hours',
-          tickFormat: function(d) {
+          tickFormat: function(d){
             return d3.format('.02f')(d);
           },
           axisLabelDistance: -10,
         },
-        forceY: [0, max_hour + 1],
+        forceY: [0, maxYValue+1],
```
jamesuejio: Add spaces between +. Just saying what Max would say, he got mad at me for not adding spaces around division.

cal/cal/static/js/app.js
```diff
@@ -180,20 +179,83 @@ function TagsDetailCtrl($scope, $http) {
...
       },
     };
-  });
+  }.bind(this);
```
jamesuejio, line 168: oh this is why my scope soup wasn't work. gj 👍

cal/api/views.py
```diff
@@ -242,4 +242,4 @@ class TagDetailEventTimeSeries(APIView):
 
     def get(self, request, *args, **kw):
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
-        return Response(tag.get_time_series(self.request.query_params.get('timezone')))
+        return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
```
jamesuejio, line 245: Is there a reason why you used self.kwargs instead of query_params? Is one better than the other? Should I change the 'timezone' to use self.kwargs? @MaxFangX I noticed you used query_params in other cases.

cal/cal/models.py
```diff
@@ -663,7 +663,7 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         return events_qs.order_by('start')
 
-    def get_time_series(self, timezone='UTC', time_step='weekly', calendar_ids=None, start=None, end=None):
+    def get_time_series(self, timezone='UTC', time_step='Week', calendar_ids=None, start=None, end=None):
```
jamesuejio: change to lower (just repeating comment here for more visibility)

cal/cal/models.py
```diff
@@ -587,7 +587,7 @@ def query(self, calendar_ids=None, start=None, end=None):
             events_qs = events_qs.filter(start__lt=datetime.now(pytz.utc))
         return events_qs.order_by('start')
 
-    def get_time_series(self, timezone='UTC', time_step='weekly', calendar_ids=None, start=None, end=None):
+    def get_time_series(self, timezone='UTC', time_step='Week', calendar_ids=None, start=None, end=None):
         """
         Returns a list of week-hour tuples corresponding to the events in this ColorCategory.
         Each week starts at the start time.
```
jamesuejio: Can you update docstring? It's not a list of week-hour tuples necessarily.

cal/cal/models.py
```diff
@@ -663,7 +663,7 @@ def query(self, calendar_ids=None, start=None, end=None):
 
         return events_qs.order_by('start')
 
-    def get_time_series(self, timezone='UTC', time_step='weekly', calendar_ids=None, start=None, end=None):
+    def get_time_series(self, timezone='UTC', time_step='Week', calendar_ids=None, start=None, end=None):
         """
         Returns a list of week-hour tuples corresponding to the events in this Tag.
         """
```
jamesuejio, line 670: See above comment regarding docstring.

cal/cal/static/js/app.js
```diff
@@ -397,20 +429,84 @@ function CategoriesDetailCtrl($scope, $http){
         xAxis: {
           axisLabel: 'Date',
           tickFormat: function(d) {
-            return d3.time.format('%m/%d/%y')(d)
-          }
+                          return d3.time.format('%m/%d/%y')(d)
+                      }
```
jamesuejio: Change indentation back

cal/cal/static/js/app.js
```diff
@@ -397,20 +429,84 @@ function CategoriesDetailCtrl($scope, $http){
...
           axisLabel: 'Hours',
-          tickFormat: function(d) {
+          tickFormat: function(d){
             return d3.format('.02f')(d);
           },
-          axisLabelDistance: -10,
+          axisLabelDistance: -10
         },
-        forceY: [0, max_hour + 1],
+        forceY: [0, maxYValue+1],
```
jamesuejio: Add spaces

cal/cal/static/templates/categoryDetails.html
```diff
@@ -1,8 +1,11 @@
 <div>
-  <h3 class="well">Average hours per week: {{ categoryHours }}<h3>
-  <nvd3 options="categoryLine" data="categoryDetails"></nvd3>
+  <h3 class="well">Average hours per {{$ctrl.timeStep}}: {{ $ctrl.averageHours }}<h3>
```
jamesuejio, line 2: beautiful

cal/cal/static/templates/tagDetails.html
```diff
@@ -1,8 +1,11 @@
...
-  <h3 class="well">Average hours per week: {{ tagHours }}<h3>
-  <nvd3 options="tagLine" data="tagDetails"></nvd3>
+  <h3 class="well">Average hours per {{$ctrl.timeStep}}: {{ $ctrl.averageHours }}<h3>
+  <button ng-click='$ctrl.showDaily()'>Daily</buton>
+  <button ng-click='$ctrl.showWeekly()'>Weekly</button>
+  <button ng-click='$ctrl.showMonthly()'>Monthly</button>
+  <nvd3 options="$ctrl.tagLine" data="ctrlDetails"></nvd3>
   <h3 class="well">Contained Events</h3>
-  <ul class="list-group" ng-repeat="tag in tagEvents">
+  <ul class="list-group" ng-repeat="tag in $ctrl.tagEvents">
```
jamesuejio, line 11: ty for fixing scope issues

jamesuejio (changes requested):

jalagar:
Tag graphs and buttons don't work for me T_T

cal/cal/helpers.py
cindytung: I'm getting these parameters from the url, so when a request is for "eventWeek"/"eventDay"/"eventMonth" I can use regex and get the time_step parameters that way. Alternatively, if I use lower case parameters, the logic for the urlpatterns and the view method would be more complicated (a lot more if/else statements i.e. if "eventWeek": time_step = "weekly" and etc). I went for simplicity...lmk what you think

jamesuejio: Ok I mean looks fine to me. @MaxFangX I think it should be fine, sometimes simplicity is better than PEP 8 :3

cal/api/views.py
cindytung: I believeeeee using query_params requires using parameters with Http requests. Using kwargs is using regex to get the parameters from the actual url string itself. It wouldn't make sense for an url to have timezone stuff in there so you should be using query_params to get the timezone

jamesuejio: SO SMART that makes a lot of sense

jamesuejio: Wait can't you just do .lower() here to convert to lower case to send to the function?

cindytung: YOU'RE SO RIGHT this is why you're cs61a head ta 👍

jamesuejio: you're a rollercoaster. you go from saying sorry i overestimated you to pulling out the head ta card

cal/cal/helpers.py
jamesuejio: NVM max she just gonna convert to lower. see below comments

cal/api/views.py
cindytung: I felt a little bad cause you just sent me like 5 messages about how insecure you are

cal/cal/static/js/app.js
MaxFangX: Yeah I was pretty damn pissed.

MaxFangX:
@jalagar Don't forget to remove your blocking review after you've verified your comments have been implemented!

cal/api/urls.py
```diff
@@ -15,12 +15,12 @@
     url(r'^v1/colorcategories/$', views.ColorCategoryList.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/?$', views.ColorCategoryDetail.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/events/?$', views.ColorCategoryDetailEvents.as_view()),
-    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/eventWeek/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
+    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/event(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
```
MaxFangX: You'll want a `/` right after `event`, it's weird to match on part of the string between two slashes.

cal/api/urls.py
```diff
@@ -15,12 +15,12 @@
...
     url(r'^v1/stats/$', views.StatisticList.as_view()),
     url(r'^v1/tags/$', views.TagList.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/?$', views.TagDetail.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/events/?$', views.TagDetailEvents.as_view()),
-    url(r'^v1/tags/(?P<pk>[0-9]+)/eventWeek/?$', views.TagDetailEventTimeSeries.as_view()),
+    url(r'^v1/tags/(?P<pk>[0-9]+)/event(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
```
MaxFangX: `/` after `event`

cal/cal/static/js/app.js
```diff
@@ -115,55 +115,54 @@ analyticsApp.component('tagList', {
...
-  $http({
-    method: 'GET',
-    url: eventweek + '.json',
-    params: {
-      timezone: query_timezone,
-    }
-  }).
-  success(function successCallback(data) {
-    $scope.tagHours = $scope.tagHours / data.length;
+function TimelineBaseCtrl($scope, $http) {
```
MaxFangX: `$http` defined but not used. There are a number of linter errors - I won't make comment for every one of them; you should install a javascript linter (`jslint`, `jshint`) and you can catch them. In case you find this helpful:

cal/cal/static/js/app.js
```diff
@@ -115,55 +115,54 @@ analyticsApp.component('tagList', {
...
+        x: date,
+        y: hours
       });
-    }
-    $scope.tagDetails.push({
-      values: events,      //values - represents the array of {x,y} data points
-      key: 'Tag Graph', //key  - the name of the series.
-      color: '#003057',  //color - optional: choose your own line color.
+    };
+    $scope.ctrlDetails = [];
```
MaxFangX: Why is this a list that you push to instead of just replacing the value of `ctrlDetails`?

cal/cal/static/js/app.js
MaxFangX: Okay, took me a while to figure out what's going on. So it looks like you're trying to find a way to reuse code between both Tags and ColorCategories by using `TimelineBaseCtrl` to set a function on the scope of both the TagDetail and CategoryDetail that both of these detail controllers can then uses to populate a scope variable that will be picked up by nvd3 to use in the graph.

Imagine if someone comes along later and sees this line `TimelineBaseCtrl.call(this, $scope, $http);`, which now does a bunch of stuff. They're not sure exactly why they need it, but somewhere deep inside a template file in another part of the codebase there was a reference to `ctrlDetails` which breaks if the TimelineBaseCtrl was not called.

In Angular, whenever you want to reuse functionality for your controllers, you're supposed to put that logic into a service. Try this architecture:

- Create a new `QueryService` that will represent things for any analytical component of our app that implements a query of events. i think this is a good name because both `Tag` and `ColorCategory` represent a way that you can query events on your calendars, and then use those queried queried events to run a number of different analyses like line graphs, cumulative stats, weekly stats, whatever. You may also have noticed that I've built Tags and ColorCategories to be extensible - it is very natural for us to implement arbitrary unions and intersections for these. For example, if I wanted to query for the times that I'm studying cs170 with Cindy or James, I can do `ColorCategory('schoolwork').intersection(Tag('James').union(Tag('Cindy')))`. I don't think stat is a good name because it is to general.
- Let this `QueryService` implement a function named something like `populateLineGraph`, which takes the same parameters you had but now instead of *mutating* a scope object that was passed in, return the data that you are currently assigning to `ctrlDetails`. Note: There are obviously many other cases where we could be using a service to simplify shared functionality between tags and categories but let's just only have your one function for now.
- In the TagDetails and CategoryDetails controllers, inject your new QueryService and use your new function to get the equivalent of the `ctrlDetails` data. Now you can pass this data into the `showGraph` function you had. If you need to refer to the data in the template, you can save it to a variable on the TagDetails or CategoryDetails controllers and reference it with `$ctrl.myDataVariable`. And all this will allow you to avoid scope soup.

For now I'll refrain from reviewing the rest as a lot is likely to change and I think it would be good to start anew.

MaxFangX (changes requested):
I checked out this branch and am super excited that I can see graphs by day/week/month. Intentionally not reviewing the entire PR (see my comments) but I can't wait for this to be merged in (:

cal/api/urls.py
MaxFangX: So probably something like `/v1/colorcategories/6969/event/week/`. Maybe `/6969/timeseries/week/`?

cal/cal/static/js/app.js
cindytung: yeah idk why. @jalagar? I'll try changing it rn and see if it works, but lmk @jalagar if there was a specific reason why it's done this way

jamesuejio: It's better style so if we want to add multiple lines we can just push more lines. You can change it for this purpose cuz we'll only have 1 line on detail pages.

jamesuejio (approved):

cal/cal/static/js/app.js
MaxFangX: Oh, nice. Maybe we should make the name of the variable more obvious

cal/api/urls.py
```diff
@@ -15,12 +15,12 @@
     url(r'^v1/colorcategories/$', views.ColorCategoryList.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/?$', views.ColorCategoryDetail.as_view()),
     url(r'^v1/colorcategories/(?P<pk>[0-9]+)/events/?$', views.ColorCategoryDetailEvents.as_view()),
-    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/eventWeek/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
+    url(r'^v1/colorcategories/(?P<pk>[0-9]+)/event/(?P<time_step>\w+)/?$', views.ColorCategoryDetailEventTimeSeries.as_view()),
```
MaxFangX: Sorry to make you change this again, but I don't think `/event/` is the best RESTful structure, since we're not returning events for a specific category, we're returning a time series for a specific category. What about we change it to `/v1/colorcategories/<pk>/timeseries/<time_step>/`

cal/api/urls.py
```diff
@@ -15,12 +15,12 @@
...
     url(r'^v1/stats/$', views.StatisticList.as_view()),
     url(r'^v1/tags/$', views.TagList.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/?$', views.TagDetail.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/events/?$', views.TagDetailEvents.as_view()),
-    url(r'^v1/tags/(?P<pk>[0-9]+)/eventWeek/?$', views.TagDetailEventTimeSeries.as_view()),
+    url(r'^v1/tags/(?P<pk>[0-9]+)/event/(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
```
MaxFangX: `v1/tags/<pk>/timeseries/<time_step>/`

MaxFangX:
General comment: Yes! We should use `then` for angular `$http`.

cal/cal/static/js/services.js
```diff
@@ -214,3 +214,30 @@ analyticsApp.service("CategoryService", ['$http', '$q', function($http, $q) {
   };
 
 }]);
+
+analyticsApp.service('QueryService', function() {
+  var ctrlDetails = [];
```
MaxFangX: I don't think you need this? Declare the variable inside the `populateData` function

cal/cal/static/templates/categoryDetails.html
```diff
@@ -1,8 +1,11 @@
...
+  <button ng-click='$ctrl.showDaily()'>Daily</buton>
+  <button ng-click='$ctrl.showWeekly()'>Weekly</button>
+  <button ng-click='$ctrl.showMonthly()'>Monthly</button>
+  <nvd3 options="$ctrl.categoryLine" data="$ctrl.ctrlDetails"></nvd3>
```
MaxFangX: I made a remark earlier, but making a review comment to be concrete: I think `ctrlDetails` is not a very clear name. @jalagar any suggestions?

cal/cal/static/templates/tagDetails.html
```diff
@@ -1,8 +1,11 @@
 <div>
-  <h3 class="well">Average hours per week: {{ tagHours }}<h3>
-  <nvd3 options="tagLine" data="tagDetails"></nvd3>
+  <h3 class="well">Average hours per {{$ctrl.timeStep}}: {{ $ctrl.averageHours }}<h3>
+  <button ng-click='$ctrl.showDaily()'>Daily</buton>
+  <button ng-click='$ctrl.showWeekly()'>Weekly</button>
+  <button ng-click='$ctrl.showMonthly()'>Monthly</button>
+  <nvd3 options="$ctrl.tagLine" data="$ctrl.ctrlDetails"></nvd3>
```
MaxFangX: ctrlDetails @jalagar

MaxFangX (commented):
After you implement this round of comments, will be ready for merge!

cal/cal/static/templates/categoryDetails.html
cindytung: ctrlGraphData ?? ¯\_(o_O)_/¯ ??? idk man

MaxFangX (approved):

MaxFangX:
Looks ready for merge!

MaxFangX:
🔥

PR #107, opened 2017-01-05
Trend line on details page

jamesuejio:
- Trend line on details page
- Reverse event order
- add start, end, calendar_ids to query
- limit pie graph and average hours to 2 decimal points

### DIFF
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -550,7 +550,7 @@ class ColorCategory(models.Model, EventCollection):
     def hours(self, calendar_ids=None, start=None, end=None):
         events = self.get_events(calendar_ids=calendar_ids, start=start, end=end)
 
-        return EventCollection(lambda: events).total_time() / 3600
+        return round(float(EventCollection(lambda: events).total_time()) / 3600, 2)
 
     def category_color(self):
         return get_color(self.calendar, self.color_index)['background']
@@ -620,7 +620,7 @@ class Tag(models.Model, EventCollection):
     def hours(self, calendar_ids=None, start=None, end=None):
         events = self.get_events(calendar_ids=calendar_ids, start=start, end=end)
 
-        return EventCollection(lambda: events).total_time() / 3600
+        return round(float(EventCollection(lambda: events).total_time()) / 3600, 2)
 
     def get_events(self, calendar_ids=None, start=None, end=None):
         """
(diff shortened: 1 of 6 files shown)


cal/cal/models.py
```diff
@@ -550,7 +550,7 @@ def __str__(self):
     def hours(self, calendar_ids=None, start=None, end=None):
         events = self.get_events(calendar_ids=calendar_ids, start=start, end=end)
 
-        return EventCollection(lambda: events).total_time() / 3600
+        return round(float(EventCollection(lambda: events).total_time()) / 3600, 2)
```
MaxFangX, line 556: Why do we do this?

cal/cal/models.py
```diff
@@ -620,7 +620,7 @@ def save(self, *args, **kwargs):
...
```
MaxFangX, line 626: Could you perhaps add a comment here explaining why this is needed? For me at least it was not readily apparent

cal/cal/static/js/services.js
```diff
@@ -220,24 +220,69 @@ analyticsApp.service('QueryService', function() {
...
+      color: '#FDB515',
+      strokeWidth: 3,
+    });
+
     return [ctrlDetails, maxYValue];
   };
 });
+
+// returns slope, intercept and r-square of the line
+function leastSquares(xSeries, ySeries) {
```
MaxFangX: global functions should go in `calendaranalytics.js`, but I think you should just attach this to `QueryService` instead.

MaxFangX (approved):
After you've implemented the tiny changes, please merge! This is great stuff - I'm excited. I love that you added a leastSquares function LOL

And thanks for installing the linter too ;)

cal/cal/models.py
jamesuejio: This is necessary to round to two decimal places. Before the pie chart was rounded down, but this becomes sort of confusing when you have pie charts per week and all your hours are off by 30 mins or something.

cal/cal/models.py
jamesuejio: ok got it. I did it in the javascript file too so I'll add a comment as well.

cal/cal/static/js/services.js
jamesuejio: Do you mean saying this.leastSquares = ...? Or just putting it directly in the this.populateData function.

cindytung: You can have another function in QueryService and call it by doing QueryService.leastSquares. You'll also have to make sure whatever controller is using this function takes in QueryService as a controller in its component

jamesuejio: I just copy and pasted it and stuck it in. Max said not to bind to controller just make it a helper function.

PR #116, opened 2017-01-07
Private event names

MaxFangX:
- Privacy endpoint and admin button on the home page so that devs don't have to feel weird when sharing their screen
- sounds eecs but ok

pls approve

pasted image at 2017_01_07 03_37 pm


cindytung (approved):
lmao

PR #117, opened 2017-01-07
inc-sync before home page rendering

cindytung:


cal/cal/views.py
```diff
@@ -10,13 +10,20 @@
 from oauth2client.client import OAuth2WebServerFlow, AccessTokenCredentials
 from oauth2client.django_orm import Storage
 
+from rest_framework.decorators import api_view
+
 from social.apps.django_app.utils import psa
 
 
+@api_view(('GET',))
```
MaxFangX: Are you sure you need this? Doesn't look like it serves any purpose

cindytung: I get a 'WSGIRequest' object has no attribute 'query_params' error without it

MaxFangX: You can use `request.GET.get('no_sync')`. We shouldn't mix in django REST framework in the pure Django part of our app.

MaxFangX (approved):

MaxFangX:
+2

PR #122, opened 2017-01-09
Moving Averages work

jamesuejio:


cal/cal/helpers.py
```diff
@@ -88,18 +90,22 @@ def get_time_series(model, timezone='UTC', time_step='week', calendar_ids=None,
...
+        increment = relativedelta(months=1)
     start = start.replace(hour=0, minute=0, second=0, microsecond=0)
     # Convert back to UTC
     start = start.astimezone(pytz.utc)
     # Rollover takes care of events that overlap time periods
     rollover = 0
+    i = 0
+
+    # For moving average
+    data_point = 0
```
cindytung: data_point = period_total = 0 for simplicity?

cal/cal/helpers.py
```diff
@@ -113,9 +119,34 @@ def get_time_series(model, timezone='UTC', time_step='week', calendar_ids=None,
             else:
                 total += (events[i].end - events[i].start).total_seconds() / 3600
             i += 1
+
+        # For moving average. Basically iterates back and adds average
+        data_point += 1
+        period_total += total
+        if data_point % period == 0:
+            average = float(period_total) / period
```
cindytung: can u do from __future__ import division so that you don't have to use float here and various other places in this file?

cal/cal/helpers.py
```diff
@@ -73,13 +73,15 @@ def get_time_series(model, timezone='UTC', time_step='week', calendar_ids=None,
...
     week_hours = []
+    moving_average_lst = []
     events = model.query(calendar_ids, start, end).order_by('start')
-    i = 0
     # Convert start to local time
     start = events[0].start.astimezone(pytz.timezone(timezone))
+    # Default week view even if `time_step` is off
+    increment = relativedelta(days=7)
     if time_step == 'day':
         # To indicate do nothing if Daily is passed in
```
cindytung: remove this line now?

cal/cal/helpers.py
```diff
@@ -113,9 +119,34 @@ def get_time_series(model, timezone='UTC', time_step='week', calendar_ids=None,
...
+    if data_point % period != 0:
+        average = float(period_total) / (data_point % period)
+        period_start = start - increment
+        sub_range = []
+        for _ in range(data_point % period):
+            sub_range = [(period_start, average)] + sub_range
+            period_start -= increment
+        moving_average_lst += sub_range
+
+    return (week_hours, moving_average_lst)
```
cindytung, line 151: hmm I think with logic-heavy algorithms like this, you should add documentation of what your logic so that if anyone in the future needs to change/fix something, they would know what u did

cindytung (changes requested):

cal/cal/helpers.py
jamesuejio: done

cal/cal/helpers.py
jamesuejio: I'm just going to change period to be 5.0. I don't want to have to import from future. I'll change this in other parts of the code too, I think this is more standard.

cal/cal/helpers.py
jamesuejio: kk

cal/cal/helpers.py
MaxFangX: wtf you beast I never knew you could do that in Python

cal/cal/static/js/services.js
```diff
@@ -278,6 +288,13 @@ analyticsApp.service('QueryService', function() {
       strokeWidth: 3,
     });
 
+    ctrlDetails.push({
+      values: movingAverage,
+      key: 'Moving Average Line',
```
MaxFangX: You should call it "XD Moving Average", where X is the number of days that you're averaging over. I actually think we should do a 30D moving average since the 5D one is still too sharp because we don't have enough data points. The moving average should average over a large enough period so that the line only makes small movements at each data point. Otherwise we might confuse the user.

cal/cal/helpers.py
```diff
@@ -88,18 +100,22 @@ def get_time_series(model, timezone='UTC', time_step='week', calendar_ids=None,
...
     start = start.astimezone(pytz.utc)
     # Rollover takes care of events that overlap time periods
     rollover = 0
+    i = 0
+
+    # For moving average
+    data_point = 0
+    period_total = 0
+    # Change period to change number of data points for average
+    period = 5.0
```
MaxFangX: Change this to 30. See comment below for my justification for a 30 day moving average.

MaxFangX (approved):
+1 After implementing comments can merge

MaxFangX:
High quality work. +1

cal/cal/helpers.py
cindytung: that's offensive

cindytung (approved):

cal/cal/static/js/services.js
```diff
@@ -238,10 +238,11 @@ analyticsApp.service('QueryService', function() {
     var ctrlDetails = [];
     var maxYValue = 0;
     var events = [];
+    var movingAverage = [];
     var xLabels = [];
     var yLabels = [];
-    for (var i = 0; i < data.length; i++) {
-      var event = data[i];
+    for (var i = 0; i < data[0].length; i++) {
```
cindytung, line 244: would combining this for loop and the for loop in 258 be more efficient, so that you don't have to iterate over the data twice?

cindytung (changes requested):

cal/cal/static/js/services.js
jamesuejio: oh shoot god

jamesuejio: wait no cuz data[0] and data[1] after different lengths, so I can't iterate through both.. I guess I could iterate through data[0] as the length of data[0] will always be greater than or equal to data[1]...

cindytung: you can do for (var i = 0, var j = 0; i < data.length, j < data[0].length; i++, j++)

jamesuejio: You mean i < data[0].length, j < data[1].length right? So one loop will stop while the other will keep going if they're different lengths?

cindytung (approved):

MaxFangX:
Damn y'all we so cute

cc @jamesuejio @cindytung @tiffanyqi

PR #123, opened 2017-01-10
Pie chart on Tag detail page that breaks down tags by categories

cindytung:


jamesuejio:
Can you make the colors match the calendar?

cal/cal/models.py
```diff
@@ -711,6 +711,39 @@ def get_time_series(self, timezone='UTC', time_step='week', calendar_ids=None, s
...
+            events_qs = reduce(lambda qs1, qs2: qs1 | qs2, querysets)
+
+            if start:
+                events_qs = events_qs.filter(end__gt=start)
+            if end:
+                events_qs = events_qs.filter(start__lt=end)
+            else:
+                events_qs = events_qs.filter(start__lt=datetime.now(pytz.utc))
+
+            hours = round(float(EventCollection(lambda:[e for e in events_qs]).total_time()) / 3600, 2)
```
jamesuejio: use `/ 3600.0` instead of using float

cal/api/urls.py
```diff
@@ -22,6 +22,7 @@
     url(r'^v1/tags/(?P<pk>[0-9]+)/?$', views.TagDetail.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/events/?$', views.TagDetailEvents.as_view()),
     url(r'^v1/tags/(?P<pk>[0-9]+)/timeseries/(?P<time_step>\w+)/?$', views.TagDetailEventTimeSeries.as_view()),
+    url(r'^v1/tags/(?P<pk>[0-9]+)/category-tags/?$', views.TagsByCategories.as_view()),
```
MaxFangX: Hmm, do you agree that `/by-category/` would be more intuitive?

cal/api/views.py
```diff
@@ -252,3 +252,11 @@ class TagDetailEventTimeSeries(APIView):
...
         tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
         return Response(tag.get_time_series(self.request.query_params.get('timezone'), time_step=self.kwargs['time_step']))
+
+
+class TagsByCategories(APIView):
+
+    def get(self, request, *args, **kw):
+        categories = Category.objects.filter(user=self.request.user)
+        tag = Tag.objects.get(user=self.request.user, id=self.kwargs['pk'])
+        return Response(tag.get_category_stats(categories))
```
MaxFangX: Cleaner to use the related_name here: `tag.get_category_stats(self.request.user.categories)`

cal/cal/models.py
```diff
@@ -711,6 +710,44 @@ def get_time_series(self, timezone='UTC', time_step='week', calendar_ids=None, s
...
+            events_qs = events_qs.order_by('start')
+
+            if start:
+                events_qs = events_qs.filter(end__gt=start)
+            if end:
+                events_qs = events_qs.filter(start__lt=end)
+            else:
+                events_qs = events_qs.filter(start__lt=datetime.now(pytz.utc))
+
+            hours = round(EventCollection(lambda:[e for e in events_qs]).total_time() / 3600.0, 2)
```
MaxFangX: `EventCollection` needs to have a `set` passed in. We're not currently using the intersects/unions yet but this will save us time in the future when we do.

Brief overview of levels of abstraction:
`query`: always returns a `QuerySet` for database level joins/intersections
`get_events`: always returns a `set` for Python level joins/intersections

cal/cal/models.py
```diff
@@ -711,6 +710,44 @@ def get_time_series(self, timezone='UTC', time_step='week', calendar_ids=None, s
...
+            for i in range(1, len(events_qs)):
+                if (events_qs[i-1].end - events_qs[i].start).total_seconds() > 0:
+                    # events overlap
+                    hours -= round((events_qs[i-1].end - events_qs[i].start).total_seconds() / 3600.0, 2)
```
MaxFangX: Could you have spaces around the `-` operator? My linter is yelling at me.

MaxFangX (changes requested):

cal/cal/models.py
MaxFangX: Also you can just do `lambda: events_qs` since a `QuerySet` is an `Iterable`.

cal/api/views.py
cindytung: hmm tag.get_category_stats(self.request.user.categories) gets me a 500 server error

MaxFangX (approved):

MaxFangX:
Didn't realize this was still sittin here, sorry. +2

PR #131, opened 2017-01-16
Partial fix for incremental sync

MaxFangX:
- Now requests a `RefreshToken` whenever you press "Connect Google Calendar"
- `import_calendars` now removes calendars if you cannot inc. sync them (this will affect users heavily once we turn on the inc sync flag). This includes external calendars and calendars for which you have write access.
- Added some print verbosity stuff for ppl working on fixing sync


jamesuejio (approved):
Works for all my non shared calendars

tiffanyqi:
doesn't work for me? :(

PR #136, opened 2017-01-17
Deselect calendar changes category pie

jamesuejio:
Also fixed bugs where deselecting and selecting didn't change cumulative

### DIFF
--- a/cal/cal/models.py
+++ b/cal/cal/models.py
@@ -605,8 +605,8 @@ class Category(models.Model, EventCollection):
 
     def query(self, calendar_ids=None, start=None, end=None):
         if self.calendar:  # Calendar Category
-            calendar_in_ids = calendar_ids is not None and self.calendar.calendar_id in calendar_ids
-            if calendar_ids is None or calendar_in_ids:
+            # Added so deselecting calendars returns 0 hours for category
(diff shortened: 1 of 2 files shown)


cal/cal/models.py
```diff
@@ -605,7 +605,11 @@ def get_events(self, calendar_ids=None, start=None, end=None):
 
     def query(self, calendar_ids=None, start=None, end=None):
         if self.calendar:  # Calendar Category
-            calendars = [self.calendar]
+            calendar_in_ids = calendar_ids is not None and self.calendar.calendar_id in calendar_ids
+            if calendar_ids is None or calendar_in_ids:
```
cindytung: if not calendar_ids or (calendar_ids and self.calendar.calendar_id in calendar_ids)

cindytung (changes requested):

cal/cal/models.py
jamesuejio: `not calendar_ids` will not work because if `calendar_ids` is `[]` this will still register as True. However if it's empty list vs. None, we want it to still return empty not go into the if statement... I think... let me test

jamesuejio: nvm you're right will change to that

cindytung (approved):

PR #138, opened 2017-01-20
Cache tag/cat stats for every calendar

cindytung:
Performance optimization 1: Cache tag/cat stats for every calendar and add up the calendars to create the final stat


cal/cal/static/js/services.js
```diff
@@ -44,53 +43,83 @@ analyticsApp.service("TagService", ['$http', '$q', function($http, $q) {
...
-          id: tag.id,
-          label: tag.label,
-          keywords: tag.keywords,
-          hours: tag.hours
-        });
+
+      for (var calendar in calendarTags) {
+        for (var n = 0; n < calendarTags[calendar].length; n++) {
+          var tag = calendarTags[calendar][n];
+          if (accumulatedTags.hasOwnProperty(tag.id)) {
```
MaxFangX: [hasOwnProperty should take in a string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty)

cal/cal/static/js/services.js
```diff
@@ -44,53 +43,83 @@ analyticsApp.service("TagService", ['$http', '$q', function($http, $q) {
...
+            var tagInfo = accumulatedTags[tag.id];
+            tagInfo[2] += tag.hours;
```
MaxFangX: Try to avoid using the `[2]`. I am not sure what tagInfo looks like so I can't give you the specifics

MaxFangX (changes requested):
Will bother you to discuss in person

cal/cal/static/js/services.js
```diff
@@ -44,53 +43,83 @@ analyticsApp.service("TagService", ['$http', '$q', function($http, $q) {
   this.tags = {};
 
   this.getTags = function(filterKey, start, end, calendarIds) {
+    var calendarTags = {}; // {calendarId : list of tags}
+    var accumulatedTags = {};
```
MaxFangX: Move these declarations just before their first usage, since these are all intermediate variables

cal/cal/static/js/services.js
MaxFangX: Ahh yeah, just use a dict instead of a list in the else clause

cal/cal/static/js/services.js
```diff
@@ -343,4 +344,58 @@ analyticsApp.service('QueryService', function() {
...
   }
-});
+
+  this.getDataForCalendarIds = function(type, start, end, calendarId, cache, cacheKey) {
+ //    As an example, tags are saved like this:
+ //    [
+ //     cacheKey1:
+ //       {
+ //         tagId1: {totalHours: 5, label: school, keywords: “UC Berkeley”},
+ //         tagId2: {totalHours: 5, label: music, tag.keywords: “Brandon Flowers”}
```
jamesuejio: take out tag.keywords

cal/cal/static/js/services.js
```diff
@@ -343,4 +344,58 @@ analyticsApp.service('QueryService', function() {
...
+          start: (start)? start.toISOString() : null,
+          end: (end)? end.toISOString() : null,
+          calendar_ids: JSON.stringify([calendarId]),
+          timezone: moment.tz.guess()
+        }
+      }).then(function successCallback(response) {
+        var modelData = response.data.results;
+        cache[cacheKey] = {};
+        for (var i = 0; i < modelData.length; i++) {
+          var model = modelData[i];
```
jamesuejio: Why don't you use a foreach loop? `i` is never used besides to get the model at that point

cal/cal/static/js/services.js
```diff
@@ -343,4 +344,58 @@ analyticsApp.service('QueryService', function() {
...
+          if (cache[cacheKey].hasOwnProperty(model.id.toString())) {
+            var newHours = cache[cacheKey][model.id][hours] + model.hours;
+            cache[cacheKey][model.id][hours] = newHours;
```
jamesuejio: Why don't you just do `cache[cacheKey][model.id][hours] += model.hours`?

jamesuejio: I also believe you can use dot notation instead of [] to access fields. 
So `cache[cacheKey][model.id].hours`? Just seems clearer that this is an attribute field, not a key value. I was going to say do this for model.id too, but I think it's clearer to user [] because each model.id points to an object.

jamesuejio (changes requested):
I pointed out some comments, but also I think there's a bug somewhere. Every time I go back and forward a week, all my numbers increase. Also my numbers are wrong for the previous week.

jamesuejio (approved):

cindytung:
bump

cindytung:
hi @MaxFangX  :~) merp

jamesuejio:
This is fucking awesome

cal/cal/static/js/services.js
```diff
@@ -526,4 +526,58 @@ analyticsApp.service('QueryService', ['$http', '$q', function($http, $q) {
...
+ //    [
+ //     cacheKey1:
+ //       {
+ //         tagId1: {hours: 5, label: "school", keywords: "UC Berkeley"},
+ //         tagId2: {hours: 5, label: "music", keywords: "The All-American Rejects"}
+ //       },
+ //     cacheKey2:
+ //       {
+ //         tagId1: {hours: 5, label: "school", keywords: “UC Berkeley”},
+ // 	      tagId2: {hours: 5, label: "music", keywords: “The All-American Rejects”}
```
MaxFangX: misindent

cal/cal/static/js/services.js
```diff
@@ -526,4 +526,58 @@ analyticsApp.service('QueryService', ['$http', '$q', function($http, $q) {
 
     return [slope, intercept, rSquare];
   }
+
+  this.getDataForCalendarIds = function(type, start, end, calendarId, cache, cacheKey) {
```
MaxFangX: This should be `getDataForCalendarId` (singular) since this applies to only one calendar.

MaxFangX (approved):
Had other architectural and API suggestions, but they are not really worth it. Great work! Sorry for the late review x.x

MaxFangX:
Make sure you pull from master and test locally before you merge

PR #139, opened 2017-01-20
can select entire row instead of just text when selecting/deselecting calendar

cindytung:
💖


jamesuejio (changes requested):
Doesn't seem to work for the very outer edges. Like whenever you hover around and the mouse becomes a little hand, it doesn't work. Also can you add it so when you hover it shows the mouse hand? I believe you add cursor: pointer; or something

jamesuejio (approved):

PR #147, opened 2017-01-23
Add preloader to calendar and show 404

tiffanyqi:
- add preloader to calendar
- 404 page, of lake victoria basically
- to see the 404, set debug to False in Dev


cal/cal/static/js/components/calendar.js
```diff
@@ -131,9 +142,11 @@ analyticsApp.controller('CalendarCtrl', function ($scope, $http, $q, uiCalendarC
        firstDay: 1,
        eventRender: $scope.eventRender,
        viewRender: this.viewRender,
+       loading: $scope.loading
      }
    };
-   
+  
+
```
jamesuejio: Whats with the two extra lines?

jamesuejio (changes requested):
404 doesn't seem to work for me. http://localhost:8000/fdsfsda gives regular Page Not Found error. Could be bower stuff, so maybe merge master in and lmk again :)

cal/cal/static/js/components/calendar.js
```diff
@@ -108,6 +108,16 @@ analyticsApp.controller('CalendarCtrl', function ($scope, $http, $q, uiCalendarC
      return element;
    };
    
+   $scope.loading = function(isLoading, view){
```
jamesuejio: I thought we didn't want to use $scope? @MaxFangX ?

jamesuejio (changes requested):

cal/cal/views.py
```diff
@@ -139,3 +140,9 @@ def google_auth(request):
     else:
         auth_uri = flow.step1_get_authorize_url()
         return HttpResponseRedirect(auth_uri)
+
+def handler404(request):
+    response = render_to_response('404.html', {},
```
MaxFangX: `render_to_response` [is "likely to be deprecated in the future"](https://docs.djangoproject.com/en/1.10/topics/http/shortcuts/#render-to-response)

cal/cal/static/js/components/calendar.js
MaxFangX: Correct! Bind to `this`

cal/cal/templates/404.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+{% load staticfiles %}
+
+{% block content %}
+
+<div class="container" style="text-align: center; height: 100%">
+
+    <h2>This link doesn't exist! :( Here's a beautiful waterfront. When you're ready go to back, click <a href="/">here to return to your dashboard</a></h2>.
```
MaxFangX: Our html is 2 spaces, not 4

MaxFangX (commented):
Small fixes, but otherwise LGTM

cal/cal/static/js/components/calendar.js
tiffanyqi: must have been an accident; thanks for checking! fixing

cal/cal/templates/404.html
tiffanyqi: cool, fixing

cal/cal/static/js/components/calendar.js
tiffanyqi: what about like `$scope.eventRender` right above it? or are we going to refactor those things later? changing this anyway, just wondering

tiffanyqi (commented):
Looks like some comments disappeared after my push: essentially this is what it was: 
- removed two line breaks in calendar.js
- changed `$scope.loading` to `this.loading` with a `bind(this)`
- changed from 4 spaces to 2 spaces in 404.html

cal/cal/views.py
tiffanyqi: @MaxFangX noooo, what should i do instead ;(

tiffanyqi:
@jalagar 404 does it work! If you go to settings.py and change your current env to `Debug = False`, you'll see it

tiffanyqi:
@jalagar implemented @MaxFangX 's changes

jamesuejio (approved):

PR #152, opened 2017-02-20
Fix sync

MaxFangX:
- Fixes some bugs relating to multiple users being on the same platform - they may both have the same calendar, which causes a MultipleObjectsReturned error
- Removes nextSyncToken field from `GoogleCredentials` class (and migrations)
- Adds `nextSyncToken` field to `GCalendar` class (and migrations)
- Adds a "Sync Calendars" button for non-superusers

Things you need to do:
- Reset credentials
- Press "Sync Calendars"

### DIFF
--- a/cal/cal/migrations/0047_auto_20170220_0038.py
+++ b/cal/cal/migrations/0047_auto_20170220_0038.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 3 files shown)


cal/cal/migrations/0047_auto_20170220_0038.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
jamesuejio, line 1: Was adding this migration intentional? It works when I re-run migrate but just making sure

jamesuejio (approved):
Looks good to me. Does this PR fix incremental sync? It worked for me adding one event but I know there were some issues

PR #153, opened 2017-02-20
Add toggle zero hour button

jamesuejio:


cindytung (approved):
u god

tiffanyqi (commented):
.

tiffanyqi (changes requested):
Can you move the button above the table instead of in the table? Thanks :)

jamesuejio:
@tiffanyqi should I just have the button outside like under sync calendars? Like toggle all zero or something? It doesn't make sense to have it separate between tags and categories?

tiffanyqi:
@jalagar ah that's true! I suggest moving it up to the topbar, next to "logout"

MaxFangX:
I don't see the button on the tag list. Am I trippin

jamesuejio:
@MaxFangX I think something is wrong with your javascript, it works for all of us.

MaxFangX (approved):
wfm

PR #155, opened 2017-02-21
Andrew's fix to tag and category details pages so the numbers are correct

jamesuejio:
Temporary fix to tag and category details page. Necessary for deployment


cindytung (approved):
lgtm

PR #156, opened 2017-02-27
Integrate new blue preloader, move "sync calendars" up to topbar

tiffanyqi:
- integrate new blue preloader, which loads immediately after you click on "sync calendars"
- move "sync calendars" up to topbar
- remove older calendar preloader
- calendar now scrolls

NOTE: can someone help make the following less hacky? Essentially all we need to do is fire `Pace.restart()` and `Pace.track()` right before we fire `/v1/sync?sync_all=true`, but this was all I could figure out how to do. Located in calendar.js, and thus removed the `a href` in logged_in.html that actually contained that URL. (Unless you're ok with hacky, then all power to you)
```
$('.sync-button').click(function() {
  Pace.restart();
  Pace.track(function() {
    $http({
      method: 'GET',
      url: "/v1/sync?sync_all=true"
    }).then(function() {
      window.location.reload();
    });
  });
});
```


MaxFangX (approved):

MaxFangX:
Though, the pace doesn't work with my tags.

tiffanyqi:
@MaxFangX seems like it didn't work for @jalagar either, his was stuck on 99%. how do you think we can test this in a way to ensure pace works for everyone?

jamesuejio:
oh @tiffanyqi it works for me. It doesn't seem super accurate though, sometimes it goes from 0-99% really fast, but then stays there for the last couple calendars. Not a huge deal though

